PR #22, opened 2017-04-04
Make a copy of metadata map before writing to it

### LINKED ISSUE #21: Race in metautils
I'm not sure if I use your library correctly, but I'm getting race with this code:
```
package main

import (
	"context"

	"github.com/mwitkow/go-grpc-middleware/util/metautils"
	"google.golang.org/grpc/metadata"
)

func main() {
	md := metadata.Pairs("key", "value")
	parent := metadata.NewContext(context.Background(), md)
	for i := 0; i < 1000; i++ {
		go func(parent context.Context) {
			ctx, cancel := context.WithCancel(parent)
			defer cancel()
			metautils.SetSingle(ctx, "key", "val")
		}(parent)
	}
}
```
The idea is that I receive gRPC request to service A which then calls concurrently multiple services (let's say B, C and D). I re-use parent context but I set some timeout for those requests. Connections between A and B-D are using retry logic from this repository (5 retries, 1 second timeout). So the race is in `metautils.SetSingle()` where multiple writes are performed on metadata map (storing x-retry-attempty header). Is it intended to not work concurrently or I'm doing something wrong? Above example is narrowed down to calling `metautils.SetSingle()` as it's not easy to reproduce, but I can prepare more adequate example if needed.

viru:
MD returned by `metadata.FromContext` should be immutable and writing
to it may cause races.

Fixes #21.

### DIFF
--- a/util/metautils/single_key.go
+++ b/util/metautils/single_key.go
@@ -45,9 +45,10 @@ func SetSingle(ctx context.Context, keyName string, keyValue string) context.Con
 		md = metadata.Pairs(keyName, keyValue)
 		return metadata.NewContext(ctx, md)
 	}
+	newMD := md.Copy()
 	k, v := encodeKeyValue(keyName, keyValue)
-	md[k] = []string{v}
-	return ctx // we use the same context because we modified the metadata in place.
+	newMD[k] = []string{v}
+	return metadata.NewContext(ctx, newMD)
 }
 
 func encodeKeyValue(k, v string) (string, string) {
--- a/util/metautils/single_key_test.go
+++ b/util/metautils/single_key_test.go
@@ -54,3 +54,13 @@ func TestSingleFailsReading(t *testing.T) {
 		})
 	}
 }
+
+func TestSingleImmutableMD(t *testing.T) {
+	key := "someKey"
+	value := "123456"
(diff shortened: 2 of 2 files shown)


mwitkow:
Can you instead have the Retry logic use `metautils.Copy`?

viru:
Can you elaborate? Do you mean using `metautils.Copy` in interceptors in `grpc_retry` package? Or in `perCallContext()`? I guess it can be done, but we don't operate on metadata there so it would be a few extra calls. Also `SetSingle()` would still be wrong (it violates usage of `metadata.FromContext()`). You probably have better vision how this package should work, so feel free to close this PR and implement the fix as you see it best :) For now we're good as we use a fork of this repo.

krzysztofdrys:
@mwitkow Do I get you right, that you would prefer to use `metautils.Copy`, instead of `MD.copy()` in this line: https://github.com/mwitkow/go-grpc-middleware/pull/22/files#diff-4336c4db50891610936b41d1e42ca285R48 ? So, in other words, use shallow copy instead of deep copy?

If yes, I think we can prepare this on Monday.

viru:
Fixed in d1d6a483040346423cc9dbf92b18c669447f19f7.

PR #23, opened 2017-04-04
Breaking change to Logging: Add grpc_ctxtags

mwitkow:
To support open tracing and other bits of middleware, it is useful to have the same semantics for gathering tags about a request (request fields, peer info etc.) This is usable for other places, such as Tracing information (Google tracing, OpenTracing).

That's why this PR extracts the request extraction logic from individual interceptors for logging and moves it to `grpc_ctxtags`.

Breaking changes:
 * grpc_logging no longer has Metadata in it
 * all request extraction is in `grpc_ctxtags`, which is the "mutable" (for upwards/downwards propagation)
 * all loggers are now immutable in the context

@adam-26 since you've been interested in grpc logging middleware, and this may affect you
@yifanzz for PR

Changes pending:
 * update examples
 * fixup docs


tags/context.go
```diff
@@ -0,0 +1,49 @@
...
+
+// Values returns a map of key to values.
+// Do not modify the underlying map, please use Set instead.
+func (t *Tags) Values() map[string]interface{} {
+	return t.values
+}
+
+// Extracts returns a pre-existing Tags object in the Context.
+// If the context wasn't set in one in a tag interceptor, a no-op Tag storage is returned.
+func Extract(ctx context.Context) *Tags {
```
devnev: If the context doesn't have a tags object, you silently return a new one that isn't in the context. If the user updates that tags object, the changes will silently be dropped. Does this really make sense?

tags/context.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+// Tags is the struct used for storing request tags between Context calls.
+// This object is *not* thread safe, and should be handled only in the context of the request.
+type Tags struct {
+	values map[string]interface{}
+}
+
+// Set sets the given key in the metadata tags.
+func (t *Tags) Set(key string, value interface{}) *Tags {
```
devnev: Your `Set` method both updates the map, and returns a tags object that you're expecting users to use. Given that updates work, it seems likely that users are going to call `Set` without using the return value (I can't even spot where you specify they shouldn't). That will work until they happen to be in a context where `Extract` returns a completely new tag object, at which point tags will suddenly be silently dropped.

yifanzz:
I wonder if we have a a method like `grpc_tags.ContextWithNewTags(context, newTagName, newTagValue)` and simply don't allow users to call `t.Set(...)`. This should avoid people accidentally forgetting to set the tag back into the context and makes that operation more seemless.

PR #24, opened 2017-04-05
add opentracing interceptors for streams and unary

mwitkow:
This adds support for opentracing interceptors:
http://opentracing.io/documentation/


tracing/opentracing/options.go
```diff
@@ -0,0 +1,55 @@
...
+	defaultOptions = &options{
+		filterOutFunc: nil,
+		tracer:        nil,
+	}
+)
+
+// FilterOutFunc allows users to provide a function that filters out certain methods from being traced.
+//
+// If it returns false, the given request will not be traced.
+type FilterOutFunc func(ctx context.Context, fullMethodName string) bool
```
improbable-yifan: `FilterFunc`? otherwise returning true for FilterOut feels like the method should be excluded.

tracing/opentracing/options.go
```diff
@@ -0,0 +1,55 @@
...
+
+type options struct {
+	filterOutFunc FilterOutFunc
+	tracer        opentracing.Tracer
+}
+
+func evaluateOptions(opts []Option) *options {
+	optCopy := &options{}
```
improbable-yifan: why not have the default options declared here?

tracing/opentracing/options.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	if optCopy.tracer == nil {
+		optCopy.tracer = opentracing.GlobalTracer()
+	}
+	return optCopy
+}
+
+type Option func(*options)
+
+// WithFilterOutFunc customizes the function used for deciding
```
improbable-yifan: finish your sentence

tracing/opentracing/server_interceptors.go
```diff
@@ -0,0 +1,81 @@
...
+
+func finishServerSpan(ctx context.Context, serverSpan opentracing.Span, err error) {
+	// Log context information
+	tags := grpc_ctxtags.Extract(ctx)
+	for k, v := range tags.Values() {
+		serverSpan.SetTag(k, v)
+	}
+	if err != nil {
+		ext.Error.Set(serverSpan, true)
+		serverSpan.LogFields(log.String("event", "error"), log.String("message", err.Error()))
```
improbable-yifan: is it worth having a field to explicitly indicate success?

tracing/opentracing/server_interceptors.go
```diff
@@ -0,0 +1,81 @@
...
+	serverSpan := tracer.StartSpan(
+		fullMethodName,
+		// this is magical, it attaches the new span to the parent parentSpanContext, and creates an unparented one if empty.
+		ext.RPCServerOption(parentSpanContext),
+		grpcTag,
+	)
+	return opentracing.ContextWithSpan(ctx, serverSpan), serverSpan
+}
+
+func finishServerSpan(ctx context.Context, serverSpan opentracing.Span, err error) {
```
improbable-yifan: Could you use the same method for both server interceptor and the client one?

tracing/opentracing/client_interceptors.go
```diff
@@ -0,0 +1,133 @@
...
+	clientSpan := tracer.StartSpan(
+		fullMethodName,
+		opentracing.ChildOf(parentSpanContext),
+		ext.SpanKindRPCClient,
+		grpcTag,
+	)
+	// Make sure we add this to the metadata of the call, so it gets propagated:
+	md, ok := metadata.FromContext(ctx)
+	if !ok {
+		md = metadata.New(nil)
```
improbable-yifan: I think we should have explicit middleware for ensuring metadata exists otherwise any interceptor that requires access to metadata will try to put in a empty metadata when it does not exists and makes metadata creation harder to reason about.

tracing/opentracing/client_interceptors.go
```diff
@@ -0,0 +1,133 @@
...
+	} else {
+		md = metautils.Copy(md)
+	}
+	if err := tracer.Inject(clientSpan.Context(), opentracing.HTTPHeaders, metadataTextMap(md)); err != nil {
+		grpclog.Printf("grpc_opentracing: failed serializing trace information: %v", err)
```
improbable-yifan: if this injection has failed will trace still work?

tracing/opentracing/interceptors_test.go
```diff
@@ -0,0 +1,170 @@
...
+	stream, err := s.Client.PingList(ctx, goodPing)
+	require.NoError(s.T(), err, "should not fail on establishing the stream")
+	for {
+		_, err := stream.Recv()
+		if err == io.EOF {
+			break
+		}
+		require.NoError(s.T(), err, "reading stream should not fail")
+	}
+	s.assertTracesCreated("/mwitkow.testproto.TestService/PingList")
```
improbable-yifan: any error case worth testing for?

tracing/opentracing/server_interceptors.go
mwitkow: No, the reason why errors are special is because tracers may decide to sample them at a higher rate:
https://github.com/opentracing/specification/blob/master/semantic_conventions.md

tracing/opentracing/client_interceptors.go
mwitkow: I will add a `metautil.Parented` or somethign down the road.

tracing/opentracing/client_interceptors.go
mwitkow: the "up stream" will, but no "down stream" traces will be added.

Anyway, this should not fail the request.

tracing/opentracing/options.go
mwitkow: Done.

tracing/opentracing/options.go
mwitkow: pattern everywhere else.

tracing/opentracing/options.go
mwitkow: done.

tracing/opentracing/server_interceptors.go
mwitkow: no, the server one has tags, the client one doesn't. Plus the client one has weirdness around the error returns.

tracing/opentracing/interceptors_test.go
mwitkow: Added error case.

mwitkow:
ROFL @improbable-yifan

PR #27, opened 2017-04-08
Feature/repo move

mwitkow:


mwitkow:
This is breaking on build due to upstream gRPC metadata changes. But it only affects tests.

PR #31, opened 2017-04-27
Make duration log configurable

dackroyd:
Reworking the duration log field so that it can be overridden to use alternate keys or value calculations

Default behaviour maintains the current `grpc.time_ms` duration

### DIFF
--- a/logging/logrus/DOC.md
+++ b/logging/logrus/DOC.md
@@ -31,6 +31,12 @@ var (
 ```
 
 ```go
+var DefaultDurationToField = DurationToTimeMillisField
+```
+DefaultDurationToField is the default implementation of converting request
+duration to a log field (key and value).
+
+```go
 var (
 	// JsonPBMarshaller is the marshaller used for serializing protobuf messages.
 	JsonPbMarshaller = &jsonpb.Marshaler{}
@@ -52,6 +58,21 @@ func DefaultCodeToLevel(code codes.Code) logrus.Level
 ```
 DefaultCodeToLevel is the default implementation of gRPC return codes to log
 levels for server side.
+
+#### func  DurationToDurationField
+
+```go
+func DurationToDurationField(duration time.Duration) (key string, value interface{})
+```
+DurationToDurationField uses the duration value to log the request duration.
+
+#### func  DurationToTimeMillisField
+
+```go
+func DurationToTimeMillisField(duration time.Duration) (key string, value interface{})
+```
+DurationToTimeMillisField converts the duration to milliseconds and uses the key
+`grpc.time_ms`.
 
 #### func  Extract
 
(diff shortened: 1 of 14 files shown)


adam-26:
This would be very useful 👍

logging/logrus/interceptors_test.go
```diff
@@ -214,3 +218,82 @@ func (s *LogrusLoggingSuite) TestPingList_WithCustomTags() {
 	assert.Contains(s.T(), msgs[1], `"level": "info"`, "OK error codes must be logged on info level.")
 	assert.Contains(s.T(), msgs[1], `"grpc.time_ms":`, "interceptor log statement should contain execution time")
 }
+
+func TestLogrusLoggingOverrideSuite(t *testing.T) {
```
mwitkow: do you need a new suite here?

How about just introduce an override function in tests that you can change in a particular test?

dackroyd: I've rebased these changes on top of your refactor, but have kept the new suite(s) for now pending clarification

With regards to your suggestion to use an override function, do you mean something like this, or have I misunderstood?:

```
opts := []grpc_logrus.Option{
	...
	grpc_zap.WithDurationField(func(duration time.Duration) zapcore.Field {
		if s.durationOverrideFunc != nil {
			return s.durationOverrideFunc(duration)
		}
		return grpc_zap.DefaultDurationToField(duration)
	}),
}
...
type zapServerSuite struct {
	*zapBaseSuite
	durationOverrideFunc grpc_zap.DurationToField
}
```

With the above, verifying behavior of the default options (i.e. that `grpc.time_ms` is used when not overridden) doesn't happen since the default option is always replaced, even if it ultimately resolves to the same function call

dackroyd:
Regarding my use case, we have a logging standard across all of our services, and part of that standard is that requests are logged with a consistent logging key regardless of implementation, and durations are logged in nanos

Ideally we'd also be capturing the request start time, but given that we have the duration + the time of the log event that is not essential.

dackroyd:
I signed it!

mwitkow:
Cool, makes sense. @dackroyd can you update this change and reuse the `baseSuite` and implement something similar fo rhte cleint interceptors.

dackroyd:
I _think_ I'm using the `baseSuite` as intended with the latest push? If I've misunderstood I might need some more guidance before I make further changes

I've updated this request with the implementation for the client interceptors, and have addressed the conflicts

While implementing the tests, I found a couple of issues with the defaults for the logging interceptors since the client interceptors were introduced, I've raised that in https://github.com/grpc-ecosystem/go-grpc-middleware/issues/36

dackroyd:
I've rebased again resolving the conflicts, so I think this should be ready to go. Thanks!

mwitkow:
Thanks! squash merging.

PR #33, opened 2017-05-01
Add Client interceptors for logging: logrus and zap

mwitkow:
This adds simple client-side interceptors for logging. For client side, the default mapping mostly consists of debug-level mapping. 

Changes in general: 
 * added `client_interceptor.go` to both logrus and zap, with fairly similar and simple semantics
 * added `client_interceptor_test.go` to both logrus and zap, with very similar tests to the server side
 * added `span.kind: server/client` to both logrus and zap server and client-side interceptors to be able to differentiate the log statements.

@dackroyd  @yifanzz

### DIFF
--- a/logging/logrus/server_interceptors_test.go
+++ b/logging/logrus/server_interceptors_test.go
@@ -95,7 +95,7 @@ func (s *logrusServerSuite) TestPingError_WithCustomLevels() {
 		{
 			code:  codes.Unauthenticated,
 			level: logrus.ErrorLevel,
-			msg:   "Unauthenticated is overwritten to ErrorLevel with customClientCodeToLevel override, which probably didn't work",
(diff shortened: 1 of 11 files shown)


logging/logrus/server_interceptors_test.go
```diff
@@ -94,7 +95,7 @@ func (s *logrusServerSuite) TestPingError_WithCustomLevels() {
 		{
 			code:  codes.Unauthenticated,
 			level: logrus.ErrorLevel,
-			msg:   "Unauthenticated is overwritten to ErrorLevel with customCodeToLevel override, which probably didn't work",
+			msg:   "Unauthenticated is overwritten to ErrorLevel with customClientCodeToLevel override, which probably didn't work",
```
dackroyd: Should this read `customCodeToLevel` (as it did before) since this is a server interceptor test?

PR #34, opened 2017-05-01
Fix/jitter

feliksik:
The [JitterUp function](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/e290027001ada7b4aaaefa959797407371765576/util/backoffutils/backoff.go#L20) documents: 

```
// JitterUp adds random jitter to the duration.
//
// This adds or substracts time from the duration within a given jitter fraction.
// For example for 10s and jitter 0.1, it will returna  time within [9s, 11s])
```

But the range is actually not reaching where it should be; it is only ranging from 9.5 to 10.5. 

This is a problem exactly for the reason we want jitter: create sufficient variance in delay. 

It is fixed in this PR. The first commit creates a failing testcase; the 2nd commit fixes the bug.


mwitkow:
Thank you! :)

PR #38, opened 2017-05-04
Remove metautils functions that call FromContext and NewContext

peter-edge:
See https://github.com/grpc/grpc-go/issues/1219, https://github.com/grpc/grpc-go/issues/1178, https://github.com/grpc/grpc-go/issues/1157, https://github.com/grpc/grpc-go/issues/1148

This code isn't used within the repository, and godoc imports says it isn't used anywhere else it can detect (although someone might be using it privately). With the new API changes, this is a bug waiting to happen, see https://github.com/grpc-ecosystem/grpc-opentracing/pull/29. I think removing these two functions might be the best call for now (better to have a compile-time error from a consumer than a runtime issue).


mwitkow:
Hmm, can this wait until they actually nuke it?

peter-edge:
I don't think they actually will remove the methods so that there is backwards compatibility, they really shouldn't remove them until 2.0. Note that right now, `GetValue(SetValue(ctx, "foo", "bar"), "foo") will not work, which could be a bug just waiting to happen. This is related to https://github.com/grpc/grpc-go/issues/1219

mwitkow:
Can you please resolve the conflicts and I'll merge this? :)

peter-edge:
Should be done. I had some issues with `./fixup.sh` so I did the merge of `util/metautils/DOC.md` manually - this might not be the right output.

NeoCN:
@peter-edge When will this pr be merged?  I  got compile error when use with latest gRPC code, in this commit https://github.com/grpc/grpc-go/commit/596a6acc87c73c3bc398a60ef0089976491ca060, the `NewContext` and `FromContext` was removed.

peter-edge:
It's up to @mwitkow, I'm not a maintainer

PR #39, opened 2017-05-05
Optimize return value for chained interceptors in length 0 and 1 cases

peter-edge:
This optimizes the return value for chained interceptors when the number of interceptors given is 0 or 1:

- When 0, a no-op interceptor is returned, to maintain backwards compatibility with existing library users (a no-op interceptor was already effectively returned, but with more internal code)
- When 1, the single interceptor is returned
- Else, a chained interceptor is returned


mwitkow:
Cheers :)

PR #43, opened 2017-05-08
The repo is moved *to* the new location, not from it.

feliksik:
About the README.md warning: I am confused. 

https://github.com/grpc-ecosystem/go-grpc-middleware is the new location, right? 

Should it not be "moved to" then?


mwitkow:
Fixed up in code generation PR.

PR #50, opened 2017-05-24
Response payload is not dumped

kazegusuri:
payload interceptor for both zap and logrus logger dumps request message as response.


mwitkow:
:man_facepalming: sorry about it!

Thanks for fixing!

PR #51, opened 2017-05-24
return error status proto in unary interceptor

kazegusuri:
Payload interceptor does not dump anything on error currently.
From grpc-go v1.3, status package was introduced and we can extract status proto from error. The proto also contains additional fields `datails`.
Now it is better to dump error response with proto status in paylaod interceptor.
zap for unary interceptor is only implemeneted for now.

### DIFF
--- a/logging/zap/client_interceptors_test.go
+++ b/logging/zap/client_interceptors_test.go
@@ -134,6 +134,7 @@ func TestZapClientOverrideSuite(t *testing.T) {
 	}
 	opts := []grpc_zap.Option{
 		grpc_zap.WithDurationField(grpc_zap.DurationToDurationField),
+		grpc_zap.WithStatusInfo(),
 	}
 	b := newBaseZapSuite(t)
 	b.InterceptorTestSuite.ClientOpts = []grpc.DialOption{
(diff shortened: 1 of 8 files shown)


mwitkow:
Thanks for contributing. 

Three things
a) Please move this to the normal logger interceptor (so it gets logged on completion like normal error status) and not the payload interceptor?
b) Please put this behind a `WithStatusInfo` option as an opt in
c) Please add the `Stream` handling, which should be exactly the same as this one.

kazegusuri:
> a) Please move this to the normal logger interceptor (so it gets logged on completion like normal error status) and not the payload interceptor?

The normal interceptor already logs error code and message (as `zap.Error(err)`).  The difference with using status proto is that it includes `details` field. The details is metadata such as stacktrace for the error, so it will be very verbose.
IMO, normal interceptor is for common case and payload interceptor is for debugging, so it is better to dump those metadata in payload interceptor.

mwitkow:
@kazegusuri 

>The normal interceptor already logs error code and message (as zap.Error(err)). The difference with using status proto is that it includes details field. The details is metadata such as stacktrace for the error, so it will be very verbose.

I agree, that's why I would like this to be behind a `WithStatusDetails` option that people can opt in with.

> IMO, normal interceptor is for common case and payload interceptor is for debugging, so it is better to dump those metadata in payload interceptor.

Not really. There's a lot of cases where you don't care about the whole payload of the request, but are interested in rich information as to why your requests fail. Moreover, the `Payload*` interceptors deal with messages exchanged, they do not deal with "end of request" similarly to the normal Interceptor (which logs the status of the response). As such, the normal Interceptor, which already has the "end of request" handling is a better place for this. 

The current PR implementation propose to "fit" nicely with the `Payload*` handlers because of teh 1:1 nature of Unary calls (you have the final error from the handler). However, you will see that it's much harder to do with `Stream` calls.

kazegusuri:
> The current PR implementation propose to "fit" nicely with the Payload* handlers because of teh 1:1 nature of Unary calls (you have the final error from the handler).

I got what you meant. Honestly, what I want now is all verbose information related to request/response messages including payload dump, error status, code, metadata/header as **one record**, because it's easy to follow what a request results in a response But I found it could fit only for Unary calls.  I agreed with you that it's good to implement with normal interceptor in order to fit for both Unary and Stream.

I will implement the feature as another interceptor or outside of this repository.  But for now, I change this PR to use normal interceptor with `WithStatusDetails` option. Is it acceptable?

mwitkow:
>I got what you meant. Honestly, what I want now is all verbose information related to request/response messages including payload dump, error status, code, metadata/header as one record, because it's easy to follow what a request results in a response But I found it could fit only for Unary calls. I agreed with you that it's good to implement with normal interceptor in order to fit for both Unary and Stream.

Yeah, that was precisely my original approach, and then Streams ruined everything.

> But for now, I change this PR to use normal interceptor with WithStatusDetails option. Is it acceptable?

Yes, adding a `WithStatusDetails` option to non-payload `UnaryServerInterceptor` and `StreamServerInterceptor` interceptor would be a great addition :)

Are you aware of the discussion in https://github.com/grpc/grpc-go/issues/1233 BTW?

kazegusuri:
> Are you aware of the discussion in grpc/grpc-go#1233 BTW?

Sure.

mwitkow:
@kazegusuri any update? :)

kazegusuri:
Sorry for late. I'm busy for now, so I will do around next Sunday.

logging/zap/client_interceptors_test.go
```diff
@@ -183,3 +184,46 @@ func (s *zapClientOverrideSuite) TestPingList_HasOverrides() {
...
+	assert.Error(s.T(), err, "error should be received on first message")
+	assert.Equal(s.T(), codes.InvalidArgument, grpc.Code(err), "gRPC status must be InvalidArgument")
+
+	msgs := s.getOutputJSONs()
+	require.Len(s.T(), msgs, 1, "one log statement should be logged")
+	m := msgs[0]
+
+	assert.Contains(s.T(), m, `grpc.service": "mwitkow.testproto.TestService"`, "all lines must contain service name")
+	assert.Contains(s.T(), m, `grpc.method": "PingList"`, "all lines must contain method name")
+	// assert.Contains(s.T(), m, fmt.Sprintf(`grpc.code": "%s"`, codes.Code(badPing.ErrorCodeReturned).String()), "all lines must contain method name")
```
kazegusuri: I have a question about StreamClientInterceptor. It probably has a problem that `logFinalClientLine` doesn't return final error correctly? It returns initialization error for grpc.NewClientStream.

kazegusuri:
I implemented incerceptors for status proto with `WithStatusInfo` option.
With the option, response like following:

**OK**

```
{
  "custom_tags.int": 1337,
  "custom_tags.string": "something",
  "grpc.code": "OK",
  "grpc.duration": 0.000170361,
  "grpc.method": "Ping",
  "grpc.response.status": {
    "msg": {}
  },
  "grpc.service": "mwitkow.testproto.TestService",
  "level": "info",
  "msg": "finished unary call",
  "peer.address": "127.0.0.1:43817",
  "span.kind": "server",
  "system": "grpc",
  "ts": 1497765054.364429
}
```

**error**

```
{
     "error": "rpc error: code = InvalidArgument desc = Userspace error.",
     "grpc.code": "InvalidArgument",
     "grpc.duration": 0.000087776,
     "grpc.method": "PingError",
     "grpc.response.status": {
          "msg": {
	      "code": 3,
	      message": "Userspace error."
	  }
     },
     "grpc.service": "mwitkow.testproto.TestService",
     "level": "info",
     "msg": "finished unary call",
     "peer.address": "127.0.0.1:44568",
     "span.kind": "server",
     "system": "grpc",
     "ts": 1497765518.4462175
     }
```

`grpcresponse.status` is also included for status OK, is it better to omit?

PR #54, opened 2017-06-07
Renamed github.com/Sirupsen/logrus to github.com/sirupsen/logrus

suprememoocow:
The package github.com/Sirupsen/logrus has been renamed to github.com/sirupsen/logrus. This PR updates all references to the new, lowercase name.

From https://github.com/sirupsen/logrus/blob/master/README.md#case-sensitivity:

> The organization's name was changed to lower-case--and this will not be changed back. If you are getting import conflicts due to case sensitivity, please use the lower-case import: github.com/sirupsen/logrus.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ install:
   # for logs/zap
   - go get go.uber.org/zap
   # for logs/logrus
-  - go get github.com/Sirupsen/logrus
+  - go get github.com/sirupsen/logrus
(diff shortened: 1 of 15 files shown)


mwitkow:
Thanks! :)

suprememoocow:
No problem 👍  Thanks for a great library! We're adopting it at GitLab for some of our new GRPC services and finding it super useful!

PR #57, opened 2017-06-19
simplify buildChain logic

0x04C2:
The previous logic is based on len(interceptors). However, the default
case can handle all the case. To remove the case check can make
buildChain clear.

@mwitkow

### DIFF
--- a/chain.go
+++ b/chain.go
@@ -16,27 +16,18 @@ import (
 // For example ChainUnaryServer(one, two, three) will execute one before two before three, and three
 // will see context changes of one and two.
 func ChainUnaryServer(interceptors ...grpc.UnaryServerInterceptor) grpc.UnaryServerInterceptor {
-	switch len(interceptors) {
-	case 0:
-		// do not want to return nil interceptor since this function was never defined to do so/for backwards compatibility
-		return func(ctx context.Context, req interface{}, _ *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
-			return handler(ctx, req)
-		}
-	case 1:
-		return interceptors[0]
-	default:
-		return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
-			buildChain := func(current grpc.UnaryServerInterceptor, next grpc.UnaryHandler) grpc.UnaryHandler {
-				return func(currentCtx context.Context, currentReq interface{}) (interface{}, error) {
-					return current(currentCtx, currentReq, info, next)
-				}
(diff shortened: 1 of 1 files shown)


mwitkow:
This is basically a revert of:
https://github.com/grpc-ecosystem/go-grpc-middleware/pull/39

@peter-edge what benefits did #39 give you?

peter-edge:
The benefit is that it doesn't add another level to your stack in the default case - this logic might be "simpler", but if I only have one interceptor, I don't want it to be wrapped with a function that is no-op in practice but has a bunch more logic.

0x04C2:
thank you for timely reply. 

For the more interceptors or no interceptor cases, there is always one no-op wrapper function at the first of chain. So, In my opinion, remove case conditions can make code clear, easy to understand and maintain. 

How do you think about this? @peter-edge @mwitkow

peter-edge:
I don't think this should be merged - it isn't actually that complicated, and the end result is better. There's a lot of cases where the default case of 1 is what happens, if you have something that takes interceptors as options, for example. Wrapping them in this interceptor code could lead to more bugs and worse performance.

What is the benefit of this PR? The `switch` statement is pretty simple.

0x04C2:
Alright. 

Basically, the interceptor implementation is built on wrapping the actual handler.

```
func _EchoService_Echo_Handler(srv interface{}, ctx context.Context, dec func(interface{}) error, interceptor grpc.UnaryServerInterceptor) (interface{}, error) {
	in := new(SimpleMessage)
	if err := dec(in); err != nil {
		return nil, err
	}
	if interceptor == nil {
		return srv.(EchoServiceServer).Echo(ctx, in)
	}
	info := &grpc.UnaryServerInfo{
		Server:     srv,
		FullMethod: "/grpc.gateway.examples.examplepb.EchoService/Echo",
	}
	handler := func(ctx context.Context, req interface{}) (interface{}, error) {
		return srv.(EchoServiceServer).Echo(ctx, req.(*SimpleMessage))
	}
	return interceptor(ctx, in, info, handler)
}
```

The empty no-op function is used to maintain the call context. Only one no-op function is ok.
If there is a lot of cases about using only one interceptor, why use buildChain?  `grpc.WithUnaryInterceptor` is good enough to use one interceptor. 

This PR is used to make code clear and easy to maintain. This is major benefit. I think the performance is ok for this case. 

@mwitkow

peter-edge:
I just really don't see the big issue here, I don't see the code as it as as less maintainable - to me, it's actually easier to reason about personally.

0x04C2:
@peter-edge haha. Got your point.
I close this PR and let's focus on other thing. 
Thank you for your time.

PR #62, opened 2017-06-28
Fix README formatting

philipithomas:


mwitkow:
Fixed out of band, thanks for pointing out though!

PR #72, opened 2017-07-26
tracing/opentracing: add ClientSpanTagKey for client span tag extension.

bom-d-van:
Right now it's impossible to pass extra tags when UnaryClientInterceptor/StreamClientInterceptor
starts a new span. This commit extends this functionality by introducing a special
context key for newClientSpanFromContext to retrieve tags if it exists and pass it
to tracer.StartSpan.

### DIFF
--- a/tracing/opentracing/client_interceptors.go
+++ b/tracing/opentracing/client_interceptors.go
@@ -8,7 +8,7 @@ import (
 	"sync"
 
 	"github.com/grpc-ecosystem/go-grpc-middleware/util/metautils"
-	"github.com/opentracing/opentracing-go"
+	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/opentracing/opentracing-go/ext"
 	"github.com/opentracing/opentracing-go/log"
 	"golang.org/x/net/context"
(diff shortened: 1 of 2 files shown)


mwitkow:
How is this different from https://github.com/grpc-ecosystem/go-grpc-middleware/tree/master/tags

bom-d-van:
> How is this different from https://github.com/grpc-ecosystem/go-grpc-middleware/tree/master/tags ?

grpc_ctxtags supports only server side, but this commit is intended for client side.

> Can these opentracing tags be anythign else than just string labels?

[It](https://godoc.org/github.com/opentracing/opentracing-go#Tag) doesn't have to, but that's unrelated here?

mwitkow:
Cool, makes sense on the client side.

Can you please add a helper function with "ClientAddContextTags" and make the marker struct private?

bom-d-van:
@mwitkow That sounds good! Please check out the new commit. :)

mwitkow:
Can you please squeeze in a test that adds it? You can add another method similar to `TestPing_PropagatesTraces`

bom-d-van:
Test added. :)

bom-d-van:
Haha, it's a long journey! Thank you! :D

PR #75, opened 2017-08-14
grpc_ctxtags: Tagging on client streamed requests

suprememoocow:
This change provides the option for client streaming bidirectional requests to be used to tag the context. 

When enabled, the first incoming client streamed request is used to set the context tags. We use it for streamed requests such as this one: https://gitlab.com/gitlab-org/gitaly-proto/blob/master/ssh.proto#L9

It can be configured using the `grpc_ctxtags.WithFieldExtractorForInitialReq`  option. (I would love a suggestion for a better name, if you have one)

```golang
opts := []grpc_ctxtags.Option{
	grpc_ctxtags.WithFieldExtractorForInitialReq(grpc_ctxtags.CodeGenRequestFieldExtractor),
}
```

Thanks for such a fantastically helpful library!

### DIFF
--- a/tags/interceptors.go
+++ b/tags/interceptors.go
@@ -33,7 +33,7 @@ func StreamServerInterceptor(opts ...Option) grpc.StreamServerInterceptor {
 			wrappedStream.WrappedContext = newCtx
 			return handler(srv, wrappedStream)
 		}
-		wrapped := &wrappedStream{stream, info, o, newCtx}
+		wrapped := &wrappedStream{stream, info, o, newCtx, true}
 		err := handler(srv, wrapped)
 		return err
 	}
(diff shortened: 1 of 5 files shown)


tags/interceptors.go
```diff
@@ -56,7 +57,9 @@ func (w *wrappedStream) Context() context.Context {
 func (w *wrappedStream) RecvMsg(m interface{}) error {
 	err := w.ServerStream.RecvMsg(m)
 	// We only do log fields extraction on the single-reqest of a server-side stream.
-	if !w.info.IsClientStream {
+	if !w.info.IsClientStream || w.opts.requestFieldsFromInitial && w.initial {
+		w.initial = false
```
suprememoocow: Presumably `RecvMsg` will never be called concurrently for the same stream twice at the same time. If this assumption is incorrect, this may need rework.

mwitkow: No, it is called from the same goroutine as it is reading msgs from the stream. So it is safe.

suprememoocow:
@mwitkow polite ping. Any chance you're able to proceed through another round of review on this?

mwitkow:
Apologies, was a little bit swamped!

Looks good, thanks for addressing comments :)

suprememoocow:
> Apologies, was a little bit swamped!

No worries at all! Many thanks for the review and accepting the contribution @mwitkow

PR #76, opened 2017-08-24
Make zap grpclog not Fatal on a simple Print

htdvisser:


mwitkow:
🤦‍♂️ Thanks for fixing.

Issue #74, opened 2017-08-07
Performance suggestion for chain builders


alcore:
Pardon the lackluster issue instead of a proper pull request, but I'm under time pressure and would otherwise just forget this. 

While I was rolling my own middleware for gRPC, I noticed https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/chain.go#L18

My suggestion is somewhat self-explanatory and looks as follows:

```go
func ChainUnaryServer(interceptors ...grpc.UnaryServerInterceptor) grpc.UnaryServerInterceptor {
	n := len(interceptors)

	if n > 1 {
		curI := 0
		lastI := n-1
		return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
			var chainHandler grpc.UnaryHandler
			chainHandler = func(currentCtx context.Context, currentReq interface{}) (interface{}, error) {
				if curI == lastI {
					return handler(currentCtx, currentReq)
				}
				curI++
				return interceptors[curI](currentCtx, currentReq, info, chainHandler)
			}

			return interceptors[0](ctx, req, info, chainHandler)
		}
	}

	if n == 1 {
		return interceptors[0]
	}

	// n == 0
	return func(ctx context.Context, req interface{}, _ *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
		return handler(ctx, req)
	}
}
```
Avoids a loop, n lamba constructions and n additional function calls. Adds n if conditions and n increments, but that should still be considerably cheaper. Branches are ordered by most likely occurrence - it's a chain after all, so I assume n is > 1. The built lambdas end up in the hot path, so I think a little bit of microoptimization won't hurt. Coincidentally, I think it's easier to reason about.

Admittedly I have not actually benchmarked it against the code in grpc_middleware (apologies - really low on time) but it _should_ (cough, cough) be quite a bit faster going by common sense. I have however been using this approach in a deployment - with no issues.

If someone wants to pick it up and work it into the interceptor factories, please go ahead. Otherwise I'll work on a proper PR, but that won't be sooner than in 2-3 weeks.

mwitkow:
@alcore that's a good suggestion, and benchmarks would be great :)

mwitkow:
Merged. Thanks!

PR #78, opened 2017-09-03
Refactored middleware chain builders

alcore:
As suggested in #74, along with the requested bench comparison, where:
- 'Master' is upstream master, 'Dev' is the proposed PR;
- 1, 3, 10 is the number of interceptors in the chain;
```
BenchmarkChainUnaryServerCallMaster1-8          300000000              5.65 ns/op            0 B/op          0 allocs/op
BenchmarkChainUnaryServerCallMaster3-8          10000000               197 ns/op             112 B/op        4 allocs/op
BenchmarkChainUnaryServerCallMaster10-8         3000000                560 ns/op             336 B/op        11 allocs/op
BenchmarkChainUnaryServerCallDev1-8             300000000              5.66 ns/op            0 B/op          0 allocs/op
BenchmarkChainUnaryServerCallDev3-8             20000000               116 ns/op             96 B/op         3 allocs/op
BenchmarkChainUnaryServerCallDev10-8            10000000               185 ns/op             96 B/op         3 allocs/op
```

Benchmark code for reference: https://gist.github.com/alcore/8b4aaec19fbdf791b5a4d2630346a75a
I didn't repeat it for the remaining builders since they follow the exact same flow, so the results should be analogous.

The difference is less than I expected initially. Even so, faster is an enemy of fast in the hot path ;-)

### DIFF
--- a/chain.go
+++ b/chain.go
@@ -16,27 +16,35 @@ import (
 // For example ChainUnaryServer(one, two, three) will execute one before two before three, and three
 // will see context changes of one and two.
 func ChainUnaryServer(interceptors ...grpc.UnaryServerInterceptor) grpc.UnaryServerInterceptor {
-	switch len(interceptors) {
-	case 0:
-		// do not want to return nil interceptor since this function was never defined to do so/for backwards compatibility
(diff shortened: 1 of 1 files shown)


mwitkow:
Cool, thanks for contributing!

PR #79, opened 2017-09-05
Fix typo

themalkolm:


mwitkow:
Fixed out of band, thanks for pointing out though!

PR #80, opened 2017-09-11
implement noopTags

flisky:
> If the context wasn't set in a tag interceptor, a no-op Tag storage is returned that will *not* be propagated in context.

However, I think the current implementation is always map-based. Am I right?

Sadly, Tags `struct` to `interface` breaks compatibility, any suggestion?

### DIFF
--- a/tags/context.go
+++ b/tags/context.go
@@ -10,43 +10,69 @@ var (
 	// ctxMarkerKey is the Context value marker used by *all* logging middleware.
 	// The logging middleware object must interf
 	ctxMarkerKey = &ctxMarker{}
+	// NoopTags is a trivial, minimum overhead implementation of Tags for which all operations are no-ops.
+	NoopTags = &noopTags{}
 )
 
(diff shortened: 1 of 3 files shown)


mwitkow:
Can you please elaborate as to what's the usecase here? You can skip the tags interceptors instead.

flisky:
It's used in [opentracing/server_interceptors](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/tracing/opentracing/server_interceptors.go#L71) automatically. And this change eliminates a memory allocation.

flisky:
Ping @mwitkow .

mwitkow:
I see, so it's an optimization. Fair enough (sorry I was off the grid for a wee while). Can you please rebase? :)

flisky:
Done, and glad you're back:)

mwitkow:
@domgreen snf @yifanzz FYI.

PR #82, opened 2017-09-19
Update DOC.md - syntax highlighting

theobouwman:
added Go syntax highlighting


mwitkow:
Um, the DOC.md is auto generated using:
https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/fixup.sh
that uses
https://github.com/davecheney/godoc2md

Consider contirbuting to that and I'll regenerate

PR #84, opened 2017-09-28
Wrap code with inline code block

izumin5210:
📝 
image


mwitkow:
This PR is out of date. Consider resynching and running `scripts/fixup.sh`

PR #85, opened 2017-10-01
Actually use jsonpb for payload marshaling.

jgiles:
Previously, the call to .addReflected(j.Message) was leading to the
default Go json.Marshal output for the message struct (since pb messages
do not themselves implement json.Marshaler). This meant the various
jsonpb.Marshaler config controls were not respected - in particular,
enum values were always printed as numbers instead of strings.

This change alters the capitalization of ouput field names to the jsonbp
camel-case style. The old output was governed by the "legacy" generated
json struct tags. See:
- https://github.com/golang/protobuf/issues/189
- https://github.com/golang/protobuf/issues/256


jgiles:
@mwitkow is there any particular process for contributions of this sort?

We're starting to use some of these tools at Paxos, and may be sending a few more PRs over time.

mwitkow:
Unfortunately ATM I'm the primary maintainer, and was super busy. Hopefully that period is over and I'll be able to review PR in a more timely manner.

Thanks for fixing this!

Issue #86, opened 2017-10-03
Configure logrus level for proto messages


coderjz:
We would like to be able to use the logrus logging interceptor to:
- Log all messages at info level
- Log the full json dump of the payload at debug level

I was able to configure the log level for the server messages (e.g. grpc_logrus.UnaryServerInterceptor) with grpc_logrus.WithLevels.  However, for payload messages it looks to be hardcoded [code link](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/logging/logrus/payload_interceptors.go#L131-L135).

Is there interest in adding something similar for the payload levels?  I'm open to submitting a pull request, but am not sure the best approach for where to set the payload level; some ideas: Add an option such as grpc_logrus.WithPayloadLevels?  Pass it directly into the Payload*Interceptor function?  Have it returned from the decider function? Other?

mwitkow:
Glad you're finding the payloads logger useful :)

Sure, that'd be great. Please use the options pattern. How about:
 - reusing the Options that already exist and documenting which ones would work
 - building new PayloadOptions struct?

I'm open to either.

coderjz:
We're using both the validation and logging packages and they're quite useful, thank you! :)

Note that I first did the change by using the log level of the entry and the tests were throwing panics.  Looking into it, it seems that logrus only sets the log level of the entry when you log at least once, and the entry defaults to loglevel.Panic ([link1](https://github.com/sirupsen/logrus/blob/master/entry.go#L50-L52), [link2](https://github.com/sirupsen/logrus/blob/master/entry.go#L94-L95)).  

Because of the panics, I instead use the entry's logger level.

To test logrus default levels:
```
	fmt.Println(logrus.NewEntry(logrus.StandardLogger()).Level)  //Prints panic
	fmt.Println(logrus.NewEntry(logrus.StandardLogger()).Logger.Level)  // Prints info
```

coderjz:
Just a question though, the logger's level usually refers to the minimum level that we allow to print logs, whereas the payload logger level is the level we are printing the logs.  Could this cause some conceptual confusion to users of the package?

mwitkow:
I'm just wondering whether this could be solved with one solution like #89, and make that extend to the payload logging.

coderjz:
I suspect that we may want to use a PayloadOptions type (feel it would be poor to overload the Options to include modifications to the Payload interceptors) -- let me know if you agree.

[...]

PR #87, opened 2017-10-04
Add variable to change the log level for payloads

coderjz:
Related to issue #86.

I decided on just using a single variable as I wasn't sure if we wanted to merge the existing Options with the Payload functions or not.


coderjz:
Different approach requested, closing this pull request.

PR #91, opened 2017-10-16
Add function AddFields that allows setting zap fields without using tags.

bakins:
This allows setting zap fields on the logger that persist. zap.fields ensure type safety on the fields.


mwitkow:
Can you please add a symmetric function to the Logrus one then? :)

bakins:
@mwitkow - I added `AddFields` to logrus as well.

PR #92, opened 2017-10-20
Use logrus entry level for payload 

coderjz:
As discussed in #86.

### DIFF
--- a/logging/logrus/payload_interceptors.go
+++ b/logging/logrus/payload_interceptors.go
@@ -8,10 +8,10 @@ import (
 
 	"fmt"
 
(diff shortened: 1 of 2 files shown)


mwitkow:
Sorry, I disappeared under a rock for a few months. Can you please rebase and:
 * call it out in the doc strings that the level of the entry is used
 * add a `example_test.go` example for this?

coderjz:
The project that I was working was stopped and I don't think I even have access to the Github repository that I sent the pull request from.

If there's no general demand for this should we just close it?

yashrsharma44:
As requested by @coderjz , closing this!

PR #97, opened 2017-11-05
extract tags from metadata

### LINKED ISSUE #93: grpc_ctxtags from metadata
I'm trying to log some per-request metadata.  The code below gets it done, producing:

```
grpc.code=OK grpc.method=Ping user=98765 peer.address=127.0.0.1:53878 span.kind=server system=grpc
```

I was wondering if there's a better way?

On the client side doing:
```

import "google.golang.org/grpc/metadata"

md := metadata.Pairs("user", "98765")
ctx := metadata.NewOutgoingContext(context.Background(), md)
client.Ping(ctx, &pb_testproto.PingRequest{})
```

And on the server, something like:
```
func UnaryServerMetdataTagInterceptor(fields ...string) grpc.UnaryServerInterceptor {
	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
		if ctxMd, ok := metadata.FromIncomingContext(ctx); ok {
			for _, field := range fields {
				if values, present := ctxMd[field]; present {
					tags := grpc_ctxtags.Extract(ctx)
					tags = tags.Set(field, strings.Join(values, ","))
				}
			}
		}
		return handler(ctx, req)
	}
}

myServer := grpc.NewServer(
	grpc_middleware.WithUnaryServerChain(
		grpc_ctxtags.UnaryServerInterceptor(),
		UnaryServerMetdataTagInterceptor("user"),
		grpc_logrus.UnaryServerInterceptor(logrusEntry, logrusOpts...),
	),
	...
)
```

Related, is there a reason `grpc_ctxtags. RequestFieldExtractorFunc` doesn't get access to the per-request context?  Would the addition of a `MetadataExtractorFunc` 
in `grpc_ctxtags.options` be welcomed?

graphaelli:
closes #93


mwitkow:
Cool. Sorry for being out for a while. 

Got a slight proposal here:
How about make this into  `RequestFieldExtractorWithContextFunc`, as a new Option (like you did) and a new type, and rebase all exisitng usage of the `WithFieldExtractor` to use the new type internally but drop the `ctx`?

graphaelli:
closing in favor of #189

PR #100, opened 2017-11-17
Support retrying gRPCs with chained interceptors

marcwilson-g:
As retries may cause a handler chain to be partially re-executed the
interceptor needs to be reset as the recursion unrolls.

Fix for #110 

Possibly regressed as part of 
https://github.com/grpc-ecosystem/go-grpc-middleware/commit/5d4723c22d97b439cd03632da61f4cfc6e94f8bb

Added tests to the retry_test as this case seemed retry specific, but did feel odd to test the chaining in the retry.

### DIFF
--- a/chain.go
+++ b/chain.go
@@ -31,7 +31,9 @@ func ChainUnaryServer(interceptors ...grpc.UnaryServerInterceptor) grpc.UnarySer
 					return handler(currentCtx, currentReq)
 				}
 				curI++
-				return interceptors[curI](currentCtx, currentReq, info, chainHandler)
+				resp, err := interceptors[curI](currentCtx, currentReq, info, chainHandler)
+				curI--
+				return resp, err
 			}
 
 			return interceptors[0](ctx, req, info, chainHandler)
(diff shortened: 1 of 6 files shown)


marcwilson-g:
I signed it!

changpingc:
ping @alcore can you review this? thanks

domgreen:
@marcwilson-g could you update the documentation explicit that the retry will call every interceptor after the retry when a retry happens.
Can you do this in both `retry/doc.go` and `doc.go`.

domgreen:
@yifanzz @mwitkow LGTM apart from updating docs 👍

marcwilson-g:
Back from leave today!
Have added comments in both retry/doc.go and doc.go as suggested (though it wasn't clear if I should also add comments to the .md files too as they seem to contain duplicate content to the doc.go files?

domgreen:
Hey Marc, sorry for the delay getting back to you. The if you run make it should auto generate the .MD files for you.

marcwilson-g:
@domgreen Done!

chain.go
```diff
@@ -31,7 +31,9 @@ func ChainUnaryServer(interceptors ...grpc.UnaryServerInterceptor) grpc.UnarySer
 					return handler(currentCtx, currentReq)
 				}
 				curI++
-				return interceptors[curI](currentCtx, currentReq, info, chainHandler)
+				resp, err := interceptors[curI](currentCtx, currentReq, info, chainHandler)
+				curI--
```
hxzhao527: Why sever side also need `curI--`?

PR #104, opened 2017-11-18
Log payload information in a goroutine

achew22:
This allows the request to be sent to the user as quickly as possible
without waiting for logging to finish.

### DIFF
--- a/logging/logrus/payload_interceptors.go
+++ b/logging/logrus/payload_interceptors.go
@@ -5,20 +5,21 @@ package grpc_logrus
 
 import (
 	"bytes"
-
 	"fmt"
+	"sync"
 
-	"github.com/sirupsen/logrus"
 	"github.com/golang/protobuf/jsonpb"
 	"github.com/golang/protobuf/proto"
 	"github.com/grpc-ecosystem/go-grpc-middleware/logging"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
 )
 
 var (
 	// JsonPBMarshaller is the marshaller used for serializing protobuf messages.
-	JsonPbMarshaller = &jsonpb.Marshaler{}
+	JsonPbMarshaller              = &jsonpb.Marshaler{}
+	fireAndForgetLoggingWaitGroup = &sync.WaitGroup{}
 )
 
 // PayloadUnaryServerInterceptor returns a new unary server interceptors that logs the payloads of requests.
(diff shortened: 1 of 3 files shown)


achew22:
Friendly ping

domgreen:
@achew22 would you mind fixing up the merge conflicts (just due to latest changes) and adding the functionality to the `logging/zap` package and I can  chase up one of the maintainers my side.

Running `make` locally will regenerate the `DOC.md` file so no need to fix just regenerate.

logging/logrus/payload_interceptors.go
```diff
@@ -31,12 +32,18 @@ func PayloadUnaryServerInterceptor(entry *logrus.Entry, decider grpc_logging.Ser
...
-		logProtoMessageAsJson(logEntry, req, "grpc.request.content", "server request payload logged as grpc.request.content field")
 		resp, err := handler(ctx, req)
-		if err == nil {
-			logProtoMessageAsJson(logEntry, resp, "grpc.response.content", "server response payload logged as grpc.request.content field")
-		}
+
+		fireAndForgetLoggingWaitGroup.Add(1)
+		// Perform payload logging in a goroutien to unblock the response to the client.
+		go func(resp interface{}, err error) {
+			logEntry := entry.WithFields(Extract(ctx).Data)
```
domgreen: Would want to do most of this code outside the go routine as the ctx may change as the app continues.

logging/logrus/payload_interceptors.go
```diff
@@ -31,12 +32,18 @@ func PayloadUnaryServerInterceptor(entry *logrus.Entry, decider grpc_logging.Ser
...
+			logProtoMessageAsJson(logEntry, req, "grpc.request.content", "server request payload logged as grpc.request.content field")
+			if err == nil {
+				logProtoMessageAsJson(logEntry, resp, "grpc.response.content", "server response payload logged as grpc.request.content field")
+			}
+			fireAndForgetLoggingWaitGroup.Done()
+		}(resp, err)
```
domgreen: Not sure that we need to add these args here, you should have access to them from the outer scope already.

domgreen:
Actually I dont think that the `logging/zap` needs to be updated due to `zaps` implementation you just do `logger.Sync()` before exiting the application.

However, a few things:
- what is the motivation behind adding this to a go routine (out of interest)?
- in doing this change you are altering the default behaviour of the middleware, you would want to have this off by default to avoid changing the behaviour for people currently using it
- add the same to `PayloadStreamServerInterceptor`

mwitkow:
I think all of the loggers we provide actually have asynchronous writing. PLease reopen if you address @domgreen comments.

PR #107, opened 2017-11-21
ctxlogger split out from grpc_logging

domgreen:
moving this out will make ctxlogger independent and mean you do not always have to use it with grpc_logging loggers.

It is then consumed by the grpc_logging loggers and we will depricate
methods on there in future.


logging/doc.go
```diff
@@ -7,10 +7,13 @@ grpc_logging is a "parent" package for gRPC logging middlewares
 
 General functionality of all middleware
 
-All logging middleware have an `Extract(ctx)` function that provides a request-scoped logger with gRPC-related fields
+All logging middleware have an adds request-scoped grpc_ctxtags with gRPC-related fields
```
yifanzz: you might want to reword this sentence

tags/logrus/server_interceptors.go
```diff
@@ -0,0 +1,32 @@
+package ctxlogger_logrus
+
+import (
+	"github.com/grpc-ecosystem/go-grpc-middleware"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/net/context"
+	"google.golang.org/grpc"
+)
+
+// PayloadUnaryServerInterceptor returns a new unary server interceptors that adds logrus.Entry to the context.
```
yifanzz: This doc line should be updated.

tags/logrus/server_interceptors.go
```diff
@@ -0,0 +1,32 @@
...
+// StreamServerInterceptor returns a new streaming server interceptor that adds logrus.Entry to the context.
+func StreamServerInterceptor(entry *logrus.Entry) grpc.StreamServerInterceptor {
+	return func(srv interface{}, stream grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
+		newCtx := newLoggerForCall(stream.Context(), entry)
+		wrapped := grpc_middleware.WrapServerStream(stream)
+		wrapped.WrappedContext = newCtx
+		err := handler(srv, wrapped)
+		return err
+	}
+}
```
yifanzz: These methods should probably stay in `logging/`. I think ctxlogger package shouldn't have any grpc specific code. Happy to discuss.

domgreen: Yeah, I've kind of been thinking about this one as well ... been going over it in my head abit trying to weigh up both approaches.

tags/logrus/server_interceptors.go
domgreen: Will fix or delete depending on the below 👍

PR #113, opened 2017-11-28
adding option to suppress interceptor logs

domgreen:


domgreen:
Review from @artsiukhou will look at passing the error into the function to give the developers more control of what calls we can suppress. A good example of this would be that some users may not wish to log all 404s and by passing the error in we can allow them to make this decision themselves.

domgreen:
Have inverted to be true by default after chat with @devnev ... changed method to be `WithDecider` as reads better in docs.

logging/zap/examples_test.go
```diff
@@ -74,3 +73,24 @@ func Example_HandlerUsageUnaryPing() {
 		return &pb_testproto.PingResponse{Value: ping.Value}, nil
 	}
 }
+
+func ExampleWithSuppressed() {
+	opts := []grpc_zap.Option{
+		grpc_zap.WithDecider(func(fullMethodName string, err error) bool {
+			// will not log gRPC calls that have the fullMethodName  and no errors
```
artsiukhou: Seems like comment is incorrect a little bit.

logging/zap/examples_test.go
```diff
@@ -74,3 +73,24 @@ func Example_HandlerUsageUnaryPing() {
...
+			if fullMethodName == "foo.bar.healthcheck" && err == nil {
```
artsiukhou: I think it's better to check `err == nil` before strings comparison. We call this function for each log record so it should affect performance as less as possible. And please fix this in other places as well (including documentation).

logging/common.go
```diff
@@ -17,6 +17,15 @@ func DefaultErrorToCode(err error) codes.Code {
 	return grpc.Code(err)
 }
 
+// Suppressed function defines rules for suppressing any interceptor logs
```
artsiukhou: Suppressed function -> Decider function?

logging/zap/examples_test.go
domgreen: updating so this comment is more clear ... we have switched so that by default the result is true (eg. we do log the gRPC call).

artsiukhou:
LGTM

logging/logrus/server_interceptors.go
```diff
@@ -26,20 +26,24 @@ func UnaryServerInterceptor(entry *logrus.Entry, opts ...Option) grpc.UnaryServe
...
-		level := o.levelFunc(code)
-		durField, durVal := o.durationFunc(time.Now().Sub(startTime))
-		fields := logrus.Fields{
-			"grpc.code": code.String(),
-			durField:    durVal,
-		}
-		if err != nil {
-			fields[logrus.ErrorKey] = err
+
+		if o.shouldLog(info.FullMethod, err) {
```
devnev: I think if you reverse the condition so that the shorter branch is indented the code will be more readable.

logging/logrus/server_interceptors.go
```diff
@@ -54,20 +58,25 @@ func StreamServerInterceptor(entry *logrus.Entry, opts ...Option) grpc.StreamSer
...
```
devnev: same here

PR #115, opened 2017-11-29
If ctx has a deadline it will be added as a tag on Extract

domgreen:


tags/context.go
```diff
@@ -1,13 +1,16 @@
...
+	"context"
+)
 
 type ctxMarker struct{}
 
 var (
 	// ctxMarkerKey is the Context value marker used by *all* logging middleware.
 	// The logging middleware object must interf
 	ctxMarkerKey = &ctxMarker{}
+	DeadlineTag  = "deadline"
```
mwitkow: `grpc.request.deadline_s` so it is scoped similarly to the rest of the fields in a context.

tags/doc.go
```diff
@@ -16,6 +16,8 @@ use `WithFieldExtractorForInitialReq` which will extract the tags from the first
 Note the tags will not be modified for subsequent requests, so this option only makes sense when the initial message
 establishes the meta-data for the stream.
 
+If there is a deadline present on the context it will be added to the tags on the context when they Extract(ctx) is called.
```
mwitkow: Please document the tag name and format of the deadline.

devnev:
> Please avoid the standard formatting of a Duration from Golang (which is a weird string https://golang.org/pkg/time/) , and convert it to a float of milliseconds (so that indexers of logs, such as Elastic, can reason about them).

Note that deadline is a time, not a duration. At least in golang's context API, the duration style is referred to as a timeout. Both forms have their up and downsides. If there's a field for request duration, an accompanying field with the timeout is definitely useful.

Agreed that _timeout_ should be a plain number rather than using a duration string so that logging systems can handle it.

tags/context.go
domgreen: updating it to `grpc.request.deadline` will be a string of `time.RFC3339` format.

domgreen:
Have updated it to `grpc.request.deadline` as suggested however as @devnev says the duration is a `time.Time` so am going to store it as `string` in `RFC3339` to match how we have mapped the `grpc.start_time`

testing/interceptor_suite.go
```diff
@@ -90,6 +90,7 @@ func (s *InterceptorTestSuite) ServerAddr() string {
 func (s *InterceptorTestSuite) SimpleCtx() context.Context {
 	ctx, _ := context.WithTimeout(context.TODO(), 2*time.Second)
 	return ctx
+
```
artsiukhou: Don't think you need this empty line.

domgreen:
i signed it

tags/interceptors_test.go
```diff
@@ -101,6 +100,29 @@ func (s *TaggingSuite) TestPing_WithCustomTags() {
...
+	assert.Equal(s.T(), tags["grpc.request.value"], "something", "the tags should contain key address")
+	assert.Len(s.T(), tags, 3, "the tags should contain only three values")
+}
+
+func (s *TaggingSuite) TestPing_WithNoDeadline() {
+	ctx := context.TODO()
+	resp, err := s.Client.Ping(ctx, goodPing)
+	require.NoError(s.T(), err, "there must be not be an on a successful call")
+	tags := tagsFromJson(s.T(), resp.Value)
+	assert.NotContains(s.T(), tags, "grpc.request.deadline", "the tags should contain a deadline")
```
artsiukhou: Wrong error message, do you mean `should not contain a deadline`?

tags/interceptors_test.go
```diff
@@ -101,6 +100,29 @@ func (s *TaggingSuite) TestPing_WithCustomTags() {
...
+	assert.Contains(s.T(), tags, "grpc.request.deadline", "the tags should contain a deadline")
+	assert.Contains(s.T(), tags, "peer.address", "the tags should contain key address")
+	assert.Equal(s.T(), tags["grpc.request.value"], "something", "the tags should contain key address")
+	assert.Len(s.T(), tags, 3, "the tags should contain only three values")
+}
+
+func (s *TaggingSuite) TestPing_WithNoDeadline() {
+	ctx := context.TODO()
+	resp, err := s.Client.Ping(ctx, goodPing)
+	require.NoError(s.T(), err, "there must be not be an on a successful call")
```
artsiukhou: Please, fix error message.

tags/interceptors_test.go
```diff
@@ -171,8 +194,9 @@ func (s *ClientStreamedTaggingSuite) TestPingStream_WithCustomTagsFirstRequest()
 
 		tags := tagsFromJson(s.T(), resp.Value)
 		assert.Contains(s.T(), tags, "peer.address", "the tags should contain key address")
+		assert.Contains(s.T(), tags, "grpc.request.deadline", "the tags should contain a deadline")
 		assert.Equal(s.T(), tags["grpc.request.value"], "something", "the tags should contain key address")
```
artsiukhou: `key address` in error message here looks wrong, should it be `grpc.request.value key`?

tags/interceptors_test.go
```diff
@@ -116,9 +138,10 @@ func (s *TaggingSuite) TestPingList_WithCustomTags() {
 		}
 		require.NoError(s.T(), err, "reading stream should not fail")
 		tags := tagsFromJson(s.T(), resp.Value)
+		assert.Contains(s.T(), tags, "grpc.request.deadline", "the tags should contain a deadline")
 		assert.Contains(s.T(), tags, "peer.address", "the tags should contain key address")
```
artsiukhou: Maybe not `key address` but just `an address` like in message above? Key is `peer.address`.

domgreen:
@artsiukhou  @mwitkow have updated based on feedback 👍

tags/interceptors_test.go
```diff
@@ -160,7 +189,7 @@ func (s *ClientStreamedTaggingSuite) TestPingStream_WithCustomTagsFirstRequest()
 		default:
 			err = stream.CloseSend()
 		}
-		require.NoError(s.T(), err, "there must be not be an on a successful call")
+		require.NoError(s.T(), err, "there must be not be an error on a successful call")
```
artsiukhou: Strange message.

tags/interceptors_test.go
```diff
@@ -148,7 +177,7 @@ type ClientStreamedTaggingSuite struct {
 
 func (s *ClientStreamedTaggingSuite) TestPingStream_WithCustomTagsFirstRequest() {
 	stream, err := s.Client.PingStream(s.SimpleCtx())
-	require.NoError(s.T(), err, "there must be not be an on a successful call")
+	require.NoError(s.T(), err, "there must be not be an error on a successful call")
```
artsiukhou: Strange message.

tags/interceptors_test.go
```diff
@@ -115,10 +142,12 @@ func (s *TaggingSuite) TestPingList_WithCustomTags() {
 			break
 		}
 		require.NoError(s.T(), err, "reading stream should not fail")
+
 		tags := tagsFromJson(s.T(), resp.Value)
-		assert.Contains(s.T(), tags, "peer.address", "the tags should contain key address")
-		assert.Equal(s.T(), tags["grpc.request.value"], "something", "the tags should contain key address")
-		assert.Len(s.T(), tags, 2, "the tags should contain only two values")
+		require.Len(s.T(), tags, 3, "the tags should contain only three values")
+		assert.Contains(s.T(), tags, "grpc.request.deadline", "the tags should not contain a deadline tag")
```
artsiukhou: `should contain`

mwitkow:
Cool whenever @devnev says it's ready.

domgreen:
@devnev @mwitkow have changed this so that we only get `grpc.request.deadline` if the grpc request itself has a deadline and it is added as a tag when the call comes in.
Have also kept in the feature where if a deadline is present on the context when you extract you will get the current deadline.
This will mean that the deadline for inner calls / scopes will be updated as we go through the call but the request deadline will stay constant.

mwitkow:
Lefit.

PR #114, opened 2017-11-29
adding grpc.start to logging interceptors

domgreen:
This pull request will add a `grpc.start_time` to the context logger showing when the gRPC call started. This used the [RFC3339](https://tools.ietf.org/html/rfc3339) format to store the data as a `string` in the context.

PR #117, opened 2017-12-06
Fix #116 - sub millisecond duration should now be measured by default

### LINKED ISSUE #116: Precision Error on durationToMilliseconds
Currently if the default duration function is used we lose precision on the duration if the call is sub millisecond.

Link to Playground
https://play.golang.org/p/P_oPECXyCU

domgreen:
Currently issue #116 is raised due to sub ms durations being logged as 0ms we can fix this by casting the duration to a float before we do the division.

This change will mean that if your `grpc.time_ms` is greater than one it will be represented as a whole number of ms `10` otherwise it will be displayed to 3 decimal places `0.542`.

### DIFF
--- a/logging/logrus/options_test.go
+++ b/logging/logrus/options_test.go
@@ -8,5 +8,5 @@ import (
 
 func TestDurationToTimeMillisField(t *testing.T) {
 	_, val := DurationToTimeMillisField(time.Microsecond * 100)
-	assert.Equal(t, val.(float32), float32(0.1),  "sub millisecond value should be return more precise")
+	assert.Equal(t, val.(float32), float32(0.1),  "sub millisecond values should be correct")
 }
--- a/logging/logrus/options_test.go
+++ b/logging/logrus/options_test.go
@@ -8,5 +8,5 @@ import (
 
 func TestDurationToTimeMillisField(t *testing.T) {
 	_, val := DurationToTimeMillisField(time.Microsecond * 100)
-	assert.Equal(t, val.(float32), float32(0.1),  "sub millisecond values should be correct")
+	assert.Equal(t, val.(float32), float32(0.1), "sub millisecond values should be correct")
 }
--- a/logging/logrus/options_test.go
+++ b/logging/logrus/options_test.go
@@ -1,0 +1,12 @@
+package grpc_logrus
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+	"time"
+)
+
+func TestDurationToTimeMillisField(t *testing.T) {
+	_, val := DurationToTimeMillisField(time.Microsecond * 100)
+	assert.Equal(t, val.(float32), float32(0.1),  "sub millisecond value should be return more precise")
+}
--- a/logging/zap/options_test.go
+++ b/logging/zap/options_test.go
@@ -1,0 +1,15 @@
+package grpc_zap
+
+import (
+	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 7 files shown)


logging/zap/options_test.go
```diff
@@ -0,0 +1,15 @@
...
+	"go.uber.org/zap/zapcore"
+	"math"
+	"testing"
+	"time"
+)
+
+func TestDurationToTimeMillisField(t *testing.T) {
+	val := DurationToTimeMillisField(time.Microsecond * 100)
+	assert.Equal(t, val.Type, zapcore.Float32Type, "should be a float type")
+	assert.Equal(t, math.Float32frombits(uint32(val.Integer)), float32(0.1), "sub millisecond values should be correct")
```
devnev: `math.Float32frombits(uint32(` feels like it's relying on knowledge of zap internals, is there any other way we can get the processed value?

domgreen: This was a tricky one to figure out, wasnt sure of it myself tbh but couldn't think of a way to do it without knowing how zapFlags worked.
If we had #114 merged we could probably grab it off the log line in an interceptor test and assert its not 0 therefore not testing this method directly.

logging/logrus/options_test.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+	"time"
+)
+
+func TestDurationToTimeMillisField(t *testing.T) {
+	_, val := DurationToTimeMillisField(time.Microsecond * 100)
+	assert.Equal(t, val.(float32), float32(0.1),  "sub millisecond values should be correct")
```
artsiukhou: This check is flaky I believe.

logging/zap/options_test.go
artsiukhou: This check is flaky as well.

domgreen:
The alternative for these tests now #114 is merged is to do:
```
assert.Contains(s.T(), msgs[1], "grpc.time_ms", "interceptor log statement should contain execution time")
	assert.NotZero(s.T(), msgs[1]["grpc.time_ms"].(float64), "duration of call should be greater than zero")
```
@devnev @artsiukhou @mwitkow  what do you think?

domgreen:
Have updated the PR so that it is inline with the above comments. No longer testing via the `DurationToTimeMillisField` explicitly and therefore do not have to know about the internals of the zap core encoding of `float`s.

mwitkow:
ok, the problem here is that if we do it this way, we will have really long, unreasonable floats.

A float for a millisecond counter is fine, as long as we cap it's precision to /10 or /100 of a millisecond.

domgreen:
@mwitkow Couple of options I could think of:

1. leave as is and close bug as not fixing
2. cast to a string and always have 2 decimal place precision `fmt.Sprintf("%.2f", durationToMilliseconds(duration))`
3. cast to string at 2 decimal places only if less than 1 and drop all decimal places if 1 or above. 

I do think it's valuable recording less than 0ms as seeing logs with `grpc.time_ms: 0` is a little confusing IMO.

domgreen:
@mwitkow have decided to go for option 3 this should keep existing behaviour for all people who expect a >1ms `grpc.time_ms` to be displayed and give some precision to 3dp for people whom are sub millisecond.

domgreen:
@mwitkow it is also now inline with `httpware` have a sample here to show outputs https://play.golang.org/p/SFpG5u9fWI

PR #118, opened 2017-12-08
change peer.address logrus representaion.

altaastro:


altaastro:
The opentracing specification describes "peer.address" logger tag as "ip:port"-a-like string, but current middleware realization writes to log a jsonpb.Marrshal'ed net.TCPAddr structure (net.Addr interface). So, perhaps, it is better to call net.Addr's .String()before passing into tag.Set() method.

altaastro:
q

altaastro:
I've signed CLA

altaastro:
Here's link to opentracing spec https://github.com/opentracing/specification/blob/master/semantic_conventions.md

altaastro:
Hello again. Can I have some feedback, please? Our company's project is using this middleware, but current
grpc-logrus doesn't follow the opentracings specs (in peer.Address), and because of that those logs are hard to parse with elastic.

domgreen:
My only concern would be if someone is already using the current representation but I also agree that moving to the standard is a good idea.
In #115 I am adding a representation of what is logged and the given types ... if we pull this after that I suggest we update those docs as well.

@yifanzz @mwitkow thoughts?

mwitkow:
Yea, legit. Sorry about that it was just a massive oversight.

PR #123, opened 2018-01-11
Add traceid and spanid to client spans

jutkko:


jutkko:
I signed it!

tracing/opentracing/client_interceptors.go
```diff
@@ -113,6 +114,8 @@ func newClientSpanFromContext(ctx context.Context, tracer opentracing.Tracer, fu
 	if parent := opentracing.SpanFromContext(ctx); parent != nil {
 		parentSpanCtx = parent.Context()
 	}
+	// Inject tags for the first time
+	ctx = grpc_ctxtags.NewTagsForCtx(ctx)
```
yifanzz: I think it would be better if this injected ctx tags if one already exists in context it shouldn't be implicitly creating one.

tracing/opentracing/client_interceptors.go
```diff
@@ -133,7 +138,18 @@ func newClientSpanFromContext(ctx context.Context, tracer opentracing.Tracer, fu
...
-func finishClientSpan(clientSpan opentracing.Span, err error) {
+func finishClientSpan(ctx context.Context, clientSpan opentracing.Span, err error) {
+	tags := grpc_ctxtags.Extract(ctx)
+	for k, v := range tags.Values() {
+		// Don't tag errors, log them instead.
+		if vErr, ok := v.(error); ok {
+			clientSpan.LogKV(k, vErr.Error())
+
+		} else {
+			clientSpan.SetTag(k, v)
```
yifanzz: why are you using `LogKV` for one but `SetTag` for the others?

jutkko: This logic is copied from the `finishClientSpan` for server interceptors. The comment says that it does not tag errors...

tracing/opentracing/client_interceptors.go
domgreen: I think we always want to have ctx_tags interceptor before tracing / logging etc. this would mean the other interceptor has already set up the context for tags.

jutkko: AFAIK this method is also used to create a tag inside the context for the very first time. How would we initialise a tag otherwise?

tracing/opentracing/client_interceptors.go
```diff
@@ -133,7 +138,18 @@ func newClientSpanFromContext(ctx context.Context, tracer opentracing.Tracer, fu
 	return opentracing.ContextWithSpan(ctxWithMetadata, clientSpan), clientSpan
 }
 
-func finishClientSpan(clientSpan opentracing.Span, err error) {
+func finishClientSpan(ctx context.Context, clientSpan opentracing.Span, err error) {
+	tags := grpc_ctxtags.Extract(ctx)
+	for k, v := range tags.Values() {
```
domgreen: Do we want to add all the tags on the request onto the client span?

tracing/opentracing/client_interceptors.go
domgreen: I think we dont need this as `grpc_tags` should always be the first thing in the interceptor chain and will set up tags etc. before the tracing / logging middleware is involked.

tracing/opentracing/client_interceptors.go
domgreen: do we want to do this? This will add **all** of the tags that are currently in the context onto the span.

tracing/opentracing/client_interceptors.go
jutkko: @domgreen You are right for the server side interceptor, but there isn't a client interceptor what injects the `grpc_tags` into the request context

jutkko: Discussed with @domgreen offline, will create a client interceptor for making sure that the tag map is present in the context.

tags/interceptors.go
```diff
@@ -65,7 +82,7 @@ func (w *wrappedStream) RecvMsg(m interface{}) error {
 	return err
 }
 
-func newTagsForCtx(ctx context.Context) context.Context {
+func NewTagsForCtx(ctx context.Context) context.Context {
```
domgreen: can we drop this back to being internal to the package?

tracing/opentracing/client_interceptors.go
```diff
@@ -94,7 +95,7 @@ func (s *tracedClientStream) finishClientSpan(err error) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	if !s.alreadyFinished {
-		finishClientSpan(s.clientSpan, err)
+		finishClientSpan(context.Background(), s.clientSpan, err)
```
domgreen: How come this is `Background()`?

tags/interceptors.go
jutkko: This method is now also used in the tracing interceptor https://github.com/grpc-ecosystem/go-grpc-middleware/pull/123#discussion-diff-161192910R118

tracing/opentracing/client_interceptors.go
jutkko: Changed

domgreen:
LGTM 👍

tags/interceptors.go
```diff
@@ -10,6 +10,15 @@ import (
 	"google.golang.org/grpc/peer"
 )
 
+// UnaryServerInterceptor returns a new unary server interceptors that sets the values for request tags.
```
yifanzz: update doc string pls

tags/interceptors.go
```diff
@@ -22,6 +31,14 @@ func UnaryServerInterceptor(opts ...Option) grpc.UnaryServerInterceptor {
 	}
 }
 
+// StreamClientInterceptor returns a new streaming server interceptor for OpenTracing.
```
yifanzz: This just injects in a ctx tags for opentracing, opentracing is not the only usecase.

tracing/opentracing/client_interceptors.go
```diff
@@ -26,7 +27,7 @@ func UnaryClientInterceptor(opts ...Option) grpc.UnaryClientInterceptor {
 		}
 		newCtx, clientSpan := newClientSpanFromContext(parentCtx, o.tracer, method)
 		err := invoker(newCtx, method, req, reply, cc, opts...)
-		finishClientSpan(clientSpan, err)
+		finishClientSpan(newCtx, clientSpan, err)
```
yifanzz: You'll need to attach the trace and span ids before you call `invoker` otherwise the context that is passed into the invoker will not have any of the opentracing information.

tracing/opentracing/client_interceptors.go
```diff
@@ -41,10 +42,10 @@ func StreamClientInterceptor(opts ...Option) grpc.StreamClientInterceptor {
 		newCtx, clientSpan := newClientSpanFromContext(parentCtx, o.tracer, method)
 		clientStream, err := streamer(newCtx, desc, cc, method, opts...)
 		if err != nil {
-			finishClientSpan(clientSpan, err)
+			finishClientSpan(newCtx, clientSpan, err)
```
yifanzz: You'll need to do the same here.

tracing/opentracing/client_interceptors.go
jutkko: The context will have the `traceid` and `spanid` inside the tag struct. `newClientSpanFromContext` makes sure that the values are there.

jutkko:
Discussed with @yifanzz offline, we are not going to do this until there a official way to extract traceID and spanID. This approach is too hacky.

PR #128, opened 2018-02-23
ctxloggers for zap and logrus should live with logging

domgreen:
I think it would fit much better if the ctxloggers for both zap and logrus lived under the logging sub folder rather than with the tags ... although they add tags to logger on the context the primary use is for them to be used to add fields to the context.


tags/zap/context.go
```diff
@@ -1,64 +1,20 @@
...
-	}
-	// Add grpc_ctxtags tags metadata until now.
-	fields := TagsToFields(ctx)
-	// Add zap fields added until now.
-	fields = append(fields, l.fields...)
-	return l.logger.With(fields...)
-}
-
-// TagsToFields transforms the Tags on the supplied context into zap fields.
-func TagsToFields(ctx context.Context) []zapcore.Field {
```
yifanzz: Should maintain the public API here

tags/zap/context.go
```diff
@@ -1,64 +1,20 @@
...
-	tags := grpc_ctxtags.Extract(ctx)
-	for k, v := range tags.Values() {
-		fields = append(fields, zap.Any(k, v))
-	}
-	return fields
-}
-
-// ToContext adds the zap.Logger to the context for extraction later.
-// Returning the new context that has been created.
-func ToContext(ctx context.Context, logger *zap.Logger) context.Context {
```
yifanzz: Same as above

tags/zap/context.go
```diff
@@ -1,64 +1,30 @@
...
-	for k, v := range tags.Values() {
-		fields = append(fields, zap.Any(k, v))
-	}
-	return fields
+	return ctxzap.TagsToFields(ctx)
 }
 
-// ToContext adds the zap.Logger to the context for extraction later.
-// Returning the new context that has been created.
+// Deprecated: use ctxzap.ToContext
```
yifanzz: Should keep the original docs line

PR #129, opened 2018-02-26
Add logging middleware using glog

### LINKED ISSUE #127: Support grpc_glog ?
Does it sound reasonable to support a logging implementation using glog ? We currently use glog and would like to use that for grpc logger.

tamalsaha:
Fixes #127

Some key points:
- I took the logrus implementation and morphed that into glog version.

- There is not Entry/Fields support in glog. So, I have created a structs for this in the tags/glog package. https://github.com/tamalsaha/go-grpc-middleware/blob/master/tags/glog/types.go

- I used the JsonFormatter code in logrus to write the Entry in glog. I used https://github.com/json-iterator/go for json marshaling for its high performance.

- glog has not DebugLevel. So, I write that as V = 4 Info.

- grpc-go package has a glog based logger. I had to copy that since there is not way to access that from outside. https://github.com/grpc/grpc-go/blob/master/grpclog/glogger/glogger.go


mwitkow:
@tamalsaha Please use as vanilla deps as possible, i.e. `encoding/json` so we don't add extra dependencies.

PR #138, opened 2018-04-17
Add check for docs

domgreen:
This commit adds a check for any change to the documentation that has not been correctly committed. It does this by regenerating the docs in Travis and then seeing if there are any changes due for staging in git.

Have also added a CONTRIBUTING.md file so that it is easier for new contributors to understand the flow.


scripts/docs.sh
```diff
@@ -22,6 +22,21 @@ function generate_markdown {
...
+}
+
+function check {
+    generate
+    count=$(git diff --numstat | wc -l | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')
+    echo $count
+    if [ "$count" = "0" ]; then
+        return 0
+    else
+        return 1
```
mwitkow: Can you make this `echo` useful information such as:
"Your markdown docs seem to be out of sync with the package docs. Please run `make` and consult `CONTRIBUTING.MD`"

.travis.yml
```diff
@@ -15,7 +15,8 @@ install:
   - dep ensure
 
 script:
- - make test 
+ - make test
+ - make checkdocs
```
mwitkow: Make sure the checkdocs are first.

domgreen:
Also need to add `realpath`

domgreen:
@mwitkow have made changes also note the annoying realpath change that I have added ... this is due to Travis not having realpath installed ... by the looks of things realpath is a can of worms.

mwitkow:
Legit. You don't need real path though, there's a bash one liner that does that.

domgreen:
Updated as per discussion to change to bash one liner 👍 merging

PR #139, opened 2018-04-24
pin google/protobuf version=1.0.0

### LINKED ISSUE #126: Use protobuf v1.0.0 instead of master
Since about 2 weeks ago there is a tagged release for golang/protobuf. It would probably make sense to use this as the constraint in the Gopkg file instead of master.

https://github.com/golang/protobuf/releases

srenatus:
The newline changes to `Gopkg.lock` stem from me having used a more recent `dep` version than when this was committed.

The change to which actual code of `github.com/google/protobuf` gets pulled in is [this](https://github.com/golang/protobuf/compare/1e59b77b52bf8e4b449a57e6f79f21226d571845...v1.0.0).

Fixes #126.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -15,19 +15,36 @@
 
 [[projects]]
   name = "github.com/gogo/protobuf"
-  packages = ["gogoproto","proto","protoc-gen-gogo/descriptor"]
+  packages = [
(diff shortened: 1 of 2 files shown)


bwplotka:
We don't have Gopkg anymore.

PR #142, opened 2018-05-02
WIP: Add MuxUnaryServer routing middleware

johnbelamaric:
Adds a per-RPC routing so you can have different middleware chains for different calls in the same server (for example, some that require auth and some that don't).

Thoughts and opinions?

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -1,0 +1,23 @@
+// gRPC Server Interceptor routing middleware.
+
+package grpc_middleware
+
+import (
+	"golang.org/x/net/context"
+	"google.golang.org/grpc"
+)
+
+type UnaryMux map[string]grpc.UnaryServerInterceptor
+
+// MuxUnaryServer creates a single interceptor out of a map of RPC to interceptors.
+func MuxUnaryServer(mux UnaryMux) grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		if i, ok := mux[info.FullMethod]; ok {
+			return i(ctx, req, info, handler)
+		}
+		if i, ok := mux["default"]; ok {
+			return i(ctx, req, info, handler)
+		}
+		return handler(ctx, req)
+	}
+}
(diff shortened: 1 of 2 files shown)


johnbelamaric:
@dmacthedestroyer @Evgeniy-L @ZachEddy is this insane?

dmacthedestroyer:
hmm.. it's interesting, but I'm still cautious :) Let me stew on it for a little while.

You could probably generalize the interceptor by having it take in some kind of predicate and another interceptor:
```
conditionFunc := func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo)bool {
  return info.SomeMetadataThing == SomeValue // or whatever the predicate is that you want
}
...
server := grpc.NewServer(
		grpc.UnaryInterceptor(
			grpc_middleware.ConditionalInterceptor(conditionFunc, myRealInterceptor)
		),
	)
```
This would mean that if the condition is met, the specified interceptor would get fired. The `muxUnaryServer` thing could be a specialized implementation of this.

Even that is probably not super great... I'll have to think it over.

domgreen:
Maybe I dont have enough context but would it not be simpler to have separate server's one that is specifically for auth and one that is not authed? Or are there services that are required on both in some way?

The logging packages do at the moment have `Decider` functions ... not sure if that fits the use case.

@mwitkow worth having a quick think of how we would do this more generally?

domgreen:
If it is specifically auth you are concerned about within [auth](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/auth/auth.go#L41) there is a call to your own auth func `authFunc(ctx)` which within there you could look at the context to decide wether to auth.

Another alternative if it is just auth is that each service can define `AuthFuncOverride` function.

Did you look into these options?

johnbelamaric:
@domgreen it wasn't so much about auth specifically, but just around the idea of having the middleware chain granularity be the method rather than the service level. But yeah, it may be easier to just use different services.

domgreen:
I think we should probably close this issue till we have a concrete use case, I would really be against adding things till we have a strong use-case. For auth and logging we currently have ways to do this built into the middleware that should suffice for most calls.

PR #144, opened 2018-05-11
Grpc zap loggerv2

jamisonhyatt:
This PR implements [zap](https://github.com/uber-go/zap) logging for the [grpc_log.LoggerV2](https://github.com/grpc/grpc-go/blob/0ebfac42d612337c9e7fb66cbe79fe25c6b818ca/grpclog/loggerv2.go#L30) interface.

The intention behind this PR was to avoid tying the underlying grpc logging level to the application log level. It seems reasonable to me to have the application at INFO while having the grpc libraries as ERROR.

To accomplish that, I could have made `ReplaceGrpcLoggerV2` create a logger from scratch, and set it's [AtomicLevel](https://github.com/uber-go/zap/blob/2dc8d10634f473f4dc678e48391a9dc8cf6ae6f9/level.go#L69) - or do what this PR does, and just conditionally log to the appropriate log level based on the flag.   This has the net effect of the underlying grpc logging never being more verbose than the zap logger that it's using.

I originally implemented this with an `AtomicLevel` where the `ReplaceGrpcLoggerV2` method required a `zap.Config` and ` ...zap.Option` where logging statements had no branching, and just relied on the AtomicLevel that was set by the environment variables. In the end, I chose against this because it felt like too much of a departure from the original - and it's simpler this way.

@mwitkow was the original implementer here - not sure if he has an opinion on this.


jamisonhyatt:
Just signed it the CLA...

jamisonhyatt:
Is this a timing thing - did I sign the CLA too fast?  I double checked the email address and github username seem to match.

logging/zap/grpclogger.go
```diff
@@ -8,6 +8,8 @@ import (
 
 	"go.uber.org/zap"
 	"google.golang.org/grpc/grpclog"
+	"os"
+	"strconv"
 )
 
 // ReplaceGrpcLogger sets the given zap.Logger as a gRPC-level logger.
```
domgreen: Can we add a depreciation notice to this pointing at V2

logging/zap/grpclogger.go
```diff
@@ -44,3 +46,103 @@ func (l *zapGrpcLogger) Printf(format string, args ...interface{}) {
...
+	switch logLevel {
+	case "", "ERROR", "error": // If env is unset, set level to ERROR.
+		zgl.severity = errorLevel
+	case "WARNING", "warning":
+		zgl.severity = warnLevel
+	case "INFO", "info":
+		zgl.severity = infoLevel
+	}
+
+	zgl.Logger = logger.With(zap.String("system", "grpc"), zap.Bool("grpc_log", true))
```
domgreen: can we reuse `SystemField` here as above?

logging/zap/grpclogger.go
domgreen: `// Deprecated: use ReplaceGrpcLoggerV2.`

logging/zap/grpclogger.go
```diff
@@ -44,3 +46,103 @@ func (l *zapGrpcLogger) Printf(format string, args ...interface{}) {
...
+}
+
+func (l *zapGrpcLoggerV2) Warningf(format string, args ...interface{}) {
+	if l.severity >= warnLevel {
+		l.Warn(fmt.Sprintf(format, args...))
+	}
+}
+
+func (l *zapGrpcLoggerV2) Error(args ...interface{}) {
+	l.Error(fmt.Sprint(args...))
```
domgreen: does this not need to be `l.logger.Error`?

domgreen: same comment on the rest of the funcs

logging/zap/grpclogger.go
jamisonhyatt: yep

logging/zap/grpclogger.go
jamisonhyatt: right - should have done that.

logging/zap/grpclogger.go
jamisonhyatt: it doesn't need to because the logger is promoted.

``` go
type zapGrpcLoggerV2 struct {
	*zap.Logger
	verbosity int
	severity  int
}
```

jamisonhyatt:
Updated

logging/zap/grpclogger.go
```diff
@@ -44,3 +48,102 @@ func (l *zapGrpcLogger) Printf(format string, args ...interface{}) {
...
+// This logger adheres to the grpc go environment variables GRPC_GO_LOG_VERBOSITY_LEVEL and GRPC_GO_LOG_SEVERITY_LEVEL.
+func ReplaceGrpcLoggerV2(logger *zap.Logger) {
+	zgl := &zapGrpcLoggerV2{Logger: logger.With(SystemField, zap.Bool("grpc_log", true))}
+	verbosity := os.Getenv("GRPC_GO_LOG_VERBOSITY_LEVEL")
+	if v, err := strconv.Atoi(verbosity); err == nil {
+		zgl.verbosity = v
+	}
+
+	logLevel := os.Getenv("GRPC_GO_LOG_SEVERITY_LEVEL")
+	switch logLevel {
```
enjoylife: In regards to the changing of logLevel dynamically, there is an [open  #issue](https://github.com/uber-go/zap/issues/581) requesting an option, e.g. `WithOptions(wrapCoreWithLevel(zap.WarnLevel)) // proposed option` to achieve this. 

If you follow that thread there is an [implementation in juju](https://github.com/juju/zaputil/blob/master/zapctx/zapctx.go#L68) which basically creates a `coreWithLevel`. 

If a similar approach to the juju code was used, then the pass through methods would be less "branchy" e.g. removing all the `if l.severity >= warnLevel`  and simply delegating the level checks to the already in place zap internals.

logging/zap/grpclogger.go
enjoylife: @domgreen is right,  since the embedded type's method is the same name, see this [quick stack](https://stackoverflow.com/questions/20861786/golang-struct-calling-embedded-type-methods-when-method-has-been-overloaded) for another example.

Heres a plug for using GoLand, as it shows how this method is recursive, which is another way of finding out the zap Logger is not being called.
image

kushp:
Sorry for the large bump, what's needed to get this over the finish line ? @jamisonhyatt @domgreen

kush-patel-hs:
I've opened https://github.com/grpc-ecosystem/go-grpc-middleware/pull/221 as a revived comments addressed version of this PR instead of expecting original author to want to push this through a year later 😅

PR #149, opened 2018-07-03
Zap custom jsonpb to allow optional gogo JSONPb serialization

antoniomo:
Hi,

This is a POC, just to ask if anyone would be interested on it.

Basically it allows me to do:

```
	// Replace JsonPb logger of zap with the gogo version
	grpc_zap.JsonPbMarshaler = &gateway.JSONPb{}
```

On my service, so then I can get proper payload logging with gogo types, such as *time.Time when using `gogoproto.stdtime = true`. In other words, it avoids this bug: gogo/protobuf#212

Since this is `go-grpc-middleware`, I guess as much compatibility with `gogo` as possible is desirable?

Let me know if this is interesting and if so, what should I change to make it acceptable to merge (maybe the same for Logrus logger, which would be trivial? Perhaps not import the `runtime.JSONPb` Marshaler from `grpc-gateway` to avoid that extra dependency, but have our own version?).

Thanks!

### DIFF
--- a/logging/zap/DOC.md
+++ b/logging/zap/DOC.md
@@ -31,7 +31,7 @@ Below is a JSON formatted example of a log that would be logged by the server in
 	{
 	  "level": "info",									// string  zap log levels
 	  "msg": "finished unary call",						// string  log message
-	
+
 	  "grpc.code": "OK",								// string  grpc status code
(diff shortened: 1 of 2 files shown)


antoniomo:
I signed it!

PR #157, opened 2018-08-08
make context errors check isRetriable

thisisneena:
In the current implementation, if the error is a context error from retry call, then the interceptor invokes a retry without checking `isRetriable`. This change fixes that.

### DIFF
--- a/retry/retry_test.go
+++ b/retry/retry_test.go
@@ -133,7 +133,6 @@ func (s *RetrySuite) TestUnary_FailsOnNonRetriableError() {
 	require.Error(s.T(), err, "error must occur from the failing service")
 	require.Equal(s.T(), codes.Internal, status.Code(err), "failure code must come from retrier")
 	require.EqualValues(s.T(), 1, s.srv.requestCount(), "one request should have been made")
-
 }
 
 func (s *RetrySuite) TestUnary_FailsOnNonRetriableContextError() {
--- a/retry/retry_test.go
+++ b/retry/retry_test.go
@@ -132,6 +132,16 @@ func (s *RetrySuite) TestUnary_FailsOnNonRetriableError() {
 	_, err := s.Client.Ping(s.SimpleCtx(), goodPing)
 	require.Error(s.T(), err, "error must occur from the failing service")
 	require.Equal(s.T(), codes.Internal, status.Code(err), "failure code must come from retrier")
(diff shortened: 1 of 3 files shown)


thisisneena:
I signed it!

domgreen:
This looks good is there a test we can add to ensure we do not regress?

timbunce:
H. Any news on this?

jeffwidman:
Is this bug still active?

thisisneena:
Looping back on this!! (sorry for all the delays – I had forgotten about this, and regaining all the context around the goals for this change took some time). 

Reading the code more closely, I don't think this change will work, since the way that per call timeouts are implemented are by adding a context timeout (in which case, we actually want to retry on a time out). Notably -- `TestUnary_PerCallDeadline_Succeeds` fails on this branch. 

But the specific problem I was trying to address still remains. Let's say we want to call an RPC `A` with the retry middleware. `A` calls some child process `B`, which fails with a context error. `A` will propagate `B`'s context error, and the retry middleware will retry (despite the fact that [this comment](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/retry/retry.go#L275) implies otherwise). I've added a new test to demonstrate this behaviour – the test fails with the code in master, but passes with this change. 

Does anyone have any thoughts on what the right behaviour is here? Should we be retrying on all propagated context errors? (/ am I misunderstanding the way we intend context errors to be propagated?)

thisisneena:
I've pushed a potential fix -- we could only retry on context errors if `perCallTimeout`s are used. This seems like a reasonable fix to me, though I'd love to hear from others. If everyone is on board, then I'll add some documentation to make this behaviour clear to those using / reading the code.

retry/retry_test.go
```diff
@@ -132,6 +132,16 @@ func (s *RetrySuite) TestUnary_FailsOnNonRetriableError() {
 	_, err := s.Client.Ping(s.SimpleCtx(), goodPing)
 	require.Error(s.T(), err, "error must occur from the failing service")
 	require.Equal(s.T(), codes.Internal, status.Code(err), "failure code must come from retrier")
+	require.EqualValues(s.T(), 1, s.srv.requestCount(), "one request should have been made")
+
```
johanbrandhorst: Nit: Could you please remove this newline?

johanbrandhorst:
Thanks for your contribution!

PR #161, opened 2018-08-23
Added retry on server stream call

lonnblad:
Added the same retry on a server stream call as for unary calls

### DIFF
--- a/retry/retry.go
+++ b/retry/retry.go
@@ -85,22 +85,45 @@ func StreamClientInterceptor(optFuncs ...CallOption) grpc.StreamClientIntercepto
 		if desc.ClientStreams {
 			return nil, grpc.Errorf(codes.Unimplemented, "grpc_retry: cannot retry on ClientStreams, set grpc_retry.Disable()")
 		}
-		logTrace(parentCtx, "grpc_retry attempt: %d, no backoff for this call", 0)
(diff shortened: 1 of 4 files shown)


mwitkow:
@lonnblad just to double check, this fixes the errors you would get on establishing a stream, i.e. when the stream cannot dial the backend and return it, so it addresses the `// TODO(mwitkow): Maybe dial and transport errors should be retriable?`

I wasn't sure about addressing this particular problem, as it really belongs to the transport layer of gRPC (e.g. having multiple backends you dial, or re-resolving and re-dialling), and it should have been solved using https://github.com/grpc/grpc-go/issues/133#issuecomment-186034983

Can you add a test case that this addresses?

lonnblad:
@mwitkow yes, that is what I was after, I understand your feeling, it would definitely be good to have it built in to the transportation layer.
Of course, I will add a test case as well.

lonnblad:
@mwitkow added a few test cases, I basically stop the server, wait and then restart it.
It would have been awesome to intercept the stream on creation, similar to `google.golang.org/grpc/tap`, and then inject different types of errors, but I believe grpc-go has some limitations there.

mwitkow:
Cool. Thanks for the contribution :)

lonnblad:
@mwitkow thanks for the repo! :) a question, how do you handle versions and releases?

PR #162, opened 2018-08-25
Use an interface for the wrapped server stream

veqryn:
I need the WrappedServerStream to satisfy an interface, that way other grpc interceptors/middlewares can both use and look for this interface.

My use case:
I am using a grpc interceptor/middleware that adds a "Request ID" onto each incoming grpc request.
I am chaining it with a second interceptor that logs all requests (the logrus payload logging).

I want downstream middlewares to be able to use and set the Context.  Currently, they are unable to, because the logging middleware wraps the ServerStream in its own different wrapper (`loggingServerStream`).

If this PR is accepted, we could then make `loggingServerStream` (and other custom wrappers) also implement `WrappedContextSettingServerStream`, allowing your middleware function `WrapServerStream` to be used and useful everywhere.

### DIFF
--- a/wrappers.go
+++ b/wrappers.go
@@ -8,6 +8,12 @@ import (
 	"google.golang.org/grpc"
 )
 
+// WrappedContextSettingServerStream allows setting the wrapped context of a grpc.ServerStream.
+type WrappedContextSettingServerStream interface {
+	grpc.ServerStream
+	SetContext(context.Context)
+}
+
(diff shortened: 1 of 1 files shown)


veqryn:
I signed it!

yashrsharma44:
> I need the WrappedServerStream to satisfy an interface, that way other grpc interceptors/middlewares can both use and look for this interface.

Since `v2` is refactored to satisfy a common interface for any grpc interceptors/middleware, we can close this PR 😄

PR #165, opened 2018-10-03
Add support for Go modules

F21:
This PR adds support for Go modules. The dep files (gopkg.toml and gopkg.lock) are kept in place because Go modules is currently experimental.

The Go modules files are initialized with the exact same versions of dependencies used by dep. The travis script was also updated to include newer Go versions and to test using Go modules in Go 1.11 and to use dep for versions below Go 1.11.


F21:
Closing/reopening to retrigger tests

F21:
Closing this as the `go.mod` and `go.sum` files were added recently in another commit.

Issue #120, opened 2017-12-25
Duplicate peer.address in zap logger output


ewang:
Hi,

I'm using the zap logging interceptor with the ctxtags interceptor. Running into an issue where the `peer.address` key appears twice.

Wondering if anyone else has run into this issue?

```
2017-12-24T22:46:05.822-0800	INFO	zap/server_interceptors.go:40	finished unary call	{"peer.address": "[::1]:63312", "grpc.start_time": "2017-12-24T22:46:05-08:00", "system": "grpc", "span.kind": "server", "grpc.service": "...", "grpc.method": "...", "peer.address": "[::1]:63312", "grpc.code": "OK", "grpc.time_ms": 398.6860046386719}
```

```
        ...
	logger, err := zap.NewDevelopment()
	if err != nil {
		log.Fatalf("failed to initialize zap logger: %v", err)
	}

	grpc_zap.ReplaceGrpcLogger(logger)

	kaParams := keepalive.ServerParameters{
		MaxConnectionIdle: 60 * time.Minute,
		Time:              60 * time.Minute,
	}

	s := grpc.NewServer(
		grpc.KeepaliveParams(kaParams),
		grpc_middleware.WithUnaryServerChain(
			grpc_ctxtags.UnaryServerInterceptor(),
			grpc_zap.UnaryServerInterceptor(logger),
			grpc_recovery.UnaryServerInterceptor(),
		),
	)
        ...
```

mwitkow:
Interesting. This indeed looks like a bug.

freddiesunarso:
I have experienced the same thing too with custom tags and peer address

freddiesunarso:
I have just found out that the issue only happens when the `grpc_ctxtags` interceptor comes first in the ChainServer function argument, i.e.

```
grpcServer := grpc.NewServer(
   grpc.UnaryInterceptor(grpc_middleware.ChainUnaryServer(
      grpc_ctxtags.UnaryServerInterceptor(opts...),
      grpc_zap.UnaryServerInterceptor(zapLogger),
   )),
)
```

But it does not happen otherwise:
```
grpcServer := grpc.NewServer(
   grpc.UnaryInterceptor(grpc_middleware.ChainUnaryServer(
      grpc_zap.UnaryServerInterceptor(zapLogger),      
      grpc_ctxtags.UnaryServerInterceptor(opts...),
   )),
)
```

PR #166, opened 2018-10-04
Remove duplicate peer.address field

### LINKED ISSUE #146: Duplicate peer.address in zap logger output ，bug
https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/tags/interceptors.go#L70

    func newTagsForCtx(ctx context.Context) context.Context {
	t := newTags()
	if peer, ok := peer.FromContext(ctx); ok {
		t.Set("peer.address", peer.Addr.String())
	}
	return setInContext(ctx, t)
    }

we needn't add peer.address here,should delete 

           if peer, ok := peer.FromContext(ctx); ok {
		t.Set("peer.address", peer.Addr.String())
	}

https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/logging/zap/server_interceptors.go#L30

       func UnaryServerInterceptor(logger *zap.Logger, opts ...Option) grpc.UnaryServerInterceptor {
	        o := evaluateServerOpt(opts)
	        return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler 
                grpc.UnaryHandler) (interface{}, error) {
		startTime := time.Now()

		newCtx := newLoggerForCall(ctx, logger, info.FullMethod, startTime)
                  ／／here read the peer.addr
		resp, err := handler(newCtx, req)
		if !o.shouldLog(info.FullMethod, err) {
			return resp, err
		}
		code := o.codeFunc(err)
		level := o.levelFunc(code)

		// re-extract logger from newCtx, as it may have extra fields that changed in the holder.
		ctx_zap.Extract(newCtx).Check(level, "finished unary call with code "+code.String()).Write(
			//here read the peer.addr again
                        zap.Error(err),
			zap.String("grpc.code", code.String()),
			o.durationFunc(time.Since(startTime)),
		)

		return resp, err
	}
}

bobheadxi:
Current issue - `peer.address` is duplicated in the zap logger middleware:

```
{"level":"info","ts":1538680314.2086248,"logger":"grpc","caller":"zap/server_interceptors.go:40","msg":"finished unary call with code OK","peer.address":"127.0.0.1:50180","grpc.start_time":"2018-10-04T12:11:54-07:00","system":"grpc","span.kind":"server","grpc.service":"pinpoint.PinpointCore","grpc.method":"GetStatus","peer.address":"127.0.0.1:50180","grpc.code":"OK","grpc.time":0.000100797}
```

Closes #146

### DIFF
--- a/tags/interceptors.go
+++ b/tags/interceptors.go
@@ -67,9 +67,6 @@ func (w *wrappedStream) RecvMsg(m interface{}) error {
 
 func newTagsForCtx(ctx context.Context) context.Context {
 	t := newTags()
-	if peer, ok := peer.FromContext(ctx); ok {
-		t.Set("peer.address", peer.Addr.String())
-	}
 	return setInContext(ctx, t)
 }
 
--- a/tags/interceptors.go
+++ b/tags/interceptors.go
@@ -7,7 +7,6 @@ import (
 	"github.com/grpc-ecosystem/go-grpc-middleware"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/peer"
 )
 
 // UnaryServerInterceptor returns a new unary server interceptors that sets the values for request tags.


bobheadxi:
I signed it!

efarem:
Hi @bobheadxi, thanks for the PR. I ran a quick test with zap and logrus, it looks like `peer.address` is only duplicated when using the zap interceptor. Removing these lines from the tags interceptor actually appears to remove the entry entirely, perhaps the problem lies within the zap interceptor?

freddiesunarso:
Hi, I would like to report that I have the same issue as well, not just on `peer.address` field, but also on every custom field that I inserted through `RequestFieldExtractorFunc`. Would you be able to fix this issue as well in this PR? Thanks.

bobheadxi:
@freddiesunarso are you encountering the issue only on the zap logger?

I will have a dig into it - the issues are probably related, though I'm a little busy lately and probably won't be able to get to this until sometime next week

freddiesunarso:
Thanks @bobheadxi, I haven't tried a different logger than zap.

I have just noticed that the issue with custom tag has also been reported in https://github.com/grpc-ecosystem/go-grpc-middleware/issues/120

freddiesunarso:
I have just found out that the issue only happens when the `grpc_ctxtags` interceptor comes first in the ChainServer function argument, i.e.

```
// create a gRPC server object
	grpcServer := grpc.NewServer(
		grpc.UnaryInterceptor(grpc_middleware.ChainUnaryServer(
                        grpc_ctxtags.UnaryServerInterceptor(opts...),
			grpc_zap.UnaryServerInterceptor(zapLogger),
		)),
	)
```

But it does not happen otherwise:
```
// create a gRPC server object
	grpcServer := grpc.NewServer(
		grpc.UnaryInterceptor(grpc_middleware.ChainUnaryServer(
			grpc_zap.UnaryServerInterceptor(zapLogger),
                        grpc_ctxtags.UnaryServerInterceptor(opts...),
		)),
	)
```

bobheadxi:
@efarem have you had a look at @zxfishhack's #187? it seems like it might fix the issue

bobheadxi:
closing in favour of #187 (I think that fixes it)

PR #181, opened 2018-12-13
Ratelimit interceptor

### LINKED ISSUE #134: Rate Limit Interceptor
Hi,

What guys do you think of a new server incerceptor thay would stall API calls for a given duration in order to prevent DoS / bruteforce.

Use case : delay every API call by 200 ms to prevent API DoS.

I know this can be done easily using an auth interceptor with a custom sleep-based function but it doesn't sound very good because it has nothing to do with auth.

ceshihao:
Add ratelimit interceptor, and implement a token bucket limiter using `github.com/juju/ratelimit`

Fix #134


jsindy:
Why are checkdocs failing?

ceshihao:
I rebase the changes on latest master.
CI pass :)

makkalot:
Any plans to merge this ?

EngHabu:
ping :) @domgreen ?

domgreen:
@ceshihao sorry I didnt pick this up earlier lots of other stuff going on with the world of work and open source so great to see some more people adding keeping the project going :+1: and @EngHabu Thanks for the ping.

An alternative way to do this could be to pass in the interceptor to the middleware eg.
```
ratelimit.UnaryServerInterceptor(WithRateLimiter(limiter))
```
and limiter just conforms to the limiter interface with a `Limit` method so that people can pass any type of limiter and the grpc_middleware does not have to depend on all the custom providers of that logic.

Thoughts?

ceshihao:
@domgreen Thanks for review. I refactor the codes according to your comments.

I change the `Limiter` interface to
```golang
type Limiter interface {
	Limit() bool
}
```
It makes everyone can pass his own limiter to the interceptor.

ratelimit/examples_test.go
```diff
@@ -0,0 +1,33 @@
+// Copyright 2018 Zheng Dayu. All Rights Reserved.
```
domgreen: please remove

ratelimit/options.go
```diff
@@ -0,0 +1,16 @@
+// See LICENSE for licensing terms.
```
domgreen: may as well remove all of this and let people view the LICENSE file ... when i get time will clean the rest of the codebase

ratelimit/options.go
```diff
@@ -0,0 +1,16 @@
...
+
+package ratelimit
+
+import "time"
+
+const infinityDuration time.Duration = 0x7fffffffffffffff
+
+type Option func(*rateLimiter)
+
+// WithRateLimiter customizes your limiter in the middleware
```
domgreen: can this be a full sentence ... how does it customize? It does give much info to the consumer.

ratelimit/ratelimit.go
```diff
@@ -0,0 +1,63 @@
...
+	return false
+}
+
+func emptyRatelimiter() *rateLimiter {
+	return &rateLimiter{
+		limiter: &emptyLimiter{},
+	}
+}
+
+// UnaryServerInterceptor returns a new unary server interceptors that performs request rate limit.
```
domgreen: `limit` > `limiting`

ratelimit/ratelimit.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// StreamServerInterceptor returns a new stream server interceptor that performs rate limiting on the request.
+func StreamServerInterceptor(opts ...Option) grpc.StreamServerInterceptor {
+	ratelimiter := emptyRatelimiter()
+	fmt.Println(ratelimiter.limiter.Limit())
+	for _, opt := range opts {
+		opt(ratelimiter)
+	}
+	fmt.Println(ratelimiter.limiter.Limit())
```
domgreen: remove Println

ratelimit/ratelimit.go
```diff
@@ -0,0 +1,63 @@
...
+	return func(srv interface{}, stream grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
+		fmt.Println(ratelimiter.limiter.Limit())
```
domgreen: and here

ratelimit/tokenbucket/tokenbucket.go
```diff
@@ -0,0 +1,28 @@
+// Copyright 2018 Zheng Dayu. All Rights Reserved.
+// See LICENSE for licensing terms.
```
domgreen: as above

ratelimit/ratelimit.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+}
+
+// StreamServerInterceptor returns a new stream server interceptor that performs rate limiting on the request.
+func StreamServerInterceptor(opts ...Option) grpc.StreamServerInterceptor {
+	ratelimiter := emptyRatelimiter()
+	fmt.Println(ratelimiter.limiter.Limit())
+	for _, opt := range opts {
+		opt(ratelimiter)
+	}
```
domgreen: I think we could get rid of the opts here ... from what i can see there is only a single opt ... the ratelimiter that the user is passing in.
Lets change it to be something like:
```
func UnaryServerInterceptor(limiter Limiter) grpc.UnaryServerInterceptor {
```

ratelimit/tokenbucket/tokenbucket_test.go
```diff
@@ -0,0 +1,22 @@
+package tokenbucket
```
domgreen: If we are getting the end user to pass in the rate limiter we can remove the whole tokenbucket package from this codebase and not depoend on juju ourselves.
The user would then wrap it using the Limiter interface so that we allow them to use any rate limit algorithm.

ratelimit/ratelimit.go
```diff
@@ -0,0 +1,63 @@
...
+	"fmt"
+
+	"golang.org/x/net/context"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type Limiter interface {
+	Limit() bool
```
domgreen: Add documentation on the the type and func :)

domgreen:
Just done a round of review @andyxning @makkalot @EngHabu feel free to add your own comments. :+1:

ratelimit/tokenbucket/tokenbucket_test.go
ceshihao: I think the package can be kept in order to give user an easy implementation of `Limiter`.

And also user can implement his own limiter if it is not enough.

domgreen: we can give an example without needing to import the dependency, the thinking behind passing in the rate limiter is so that we do not have to maintain the rate limiters but just wrap them to meet the Limiter interface that has been defined.

ceshihao: 👌remove it from repo

ratelimit/doc.go
```diff
@@ -0,0 +1,12 @@
+/*
+`ratelimit` a generic server-side ratelimit middleware for gRPC.
+
+Server Side Ratelimit Middleware
+
+It allows to do grpc rate limit by your own rate limiter (e.g. token bucket, leaky bucket, etc.)
+
+`ratelimit/tokenbucket`provides an implementation based on token bucket `github.com/juju/ratelimit`.
```
domgreen: Can we update ... I dont think this is true anymore :)

ratelimit/examples_test.go
```diff
@@ -0,0 +1,26 @@
+package ratelimit_test
+
+import (
+	"time"
+
+	"github.com/ceshihao/ratelimiter/tokenbucket"
```
domgreen: This is no longer in the project.

Tio do the example can we create a test rate limiter that meets the interface :+1:

ceshihao: Do you mean to add a fake limiter (e.g. `allwaysPassLimiter`) in the example instead of the one based on `juju/ratelimit`?

domgreen: yep, lets just add a fake that returns true (or get creative) but one that is not exported ... dont really want the dep just for an example.

ceshihao: Done

PR #184, opened 2019-01-04
Add if the trace was sampled to the ctxTags

domgreen:
Currently we add traceId and spanId to the ctxTags ... which eventually end up on each log line if you are using the ctxTags for logging ... this ends up in a very annoying situation where we have a log line with a span / trace id but when you go hunting for the trace you can never find it. 

A fix to this is to add the `trace.sampled` to the context as well so that we can then identify that the trace will not be in our tracing system.

Previously for a sampled trace would have had :
```
trace.traceid:123
trace.spanid:987
```
Which will now be:

```
trace.traceid:123
trace.spanid:987
trace.sampled:true
```

Previously for a trace that was *not* samped would have had :
 ```
trace.traceid:456
trace.spanid:654
```
Which will now be:

```
trace.traceid:456
trace.spanid:654
trace.sampled:false
```

### DIFF
--- a/tracing/opentracing/id_extract.go
+++ b/tracing/opentracing/id_extract.go
@@ -12,10 +12,10 @@ const (
 	TagTraceId           = "trace.traceid"
 	TagSpanId            = "trace.spanid"
 	TagSampled           = "trace.sampled"
-	JaegerNotSampledFlag = "0"
+	jaegerNotSampledFlag = "0"
 )
 
-// injectOpentracingIdsToTags writes the given context to the ctxtags if the trace.
+// injectOpentracingIdsToTags writes trace data to ctxtags.
 // This is done in an incredibly hacky way, because the public-facing interface of opentracing doesn't give access to
 // the TraceId and SpanId of the SpanContext. Only the Tracer's Inject/Extract methods know what these are.
 // Most tracers have them encoded as keys with 'traceid' and 'spanid':
@@ -35,15 +35,16 @@ type tagsCarrier struct {
 }
 
 func (t *tagsCarrier) Set(key, val string) {
-	if strings.Contains(strings.ToLower(key), "traceid") {
+	key = strings.ToLower(key)
+	if strings.Contains(key, "traceid") {
 		t.Tags.Set(TagTraceId, val) // this will most likely be base-16 (hex) encoded
 	}
 
-	if strings.Contains(strings.ToLower(key), "spanid") && !strings.Contains(strings.ToLower(key), "parent") {
+	if strings.Contains(key, "spanid") && !strings.Contains(strings.ToLower(key), "parent") {
 		t.Tags.Set(TagSpanId, val) // this will most likely be base-16 (hex) encoded
 	}
 
-	if strings.Contains(strings.ToLower(key), "sampled") {
+	if strings.Contains(key, "sampled") {
 		t.Tags.Set(TagSampled, val)
 	}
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: tracing/opentracing/id_extract.go [hunk 1]
```diff
@@ -12,39 +9,55 @@ import (
...
-type hackyTagsCarrier struct {
+// tagsCarrier is a really hacky way of
+type tagsCarrier struct {
 	grpc_ctxtags.Tags
 }
 
-func (t *hackyTagsCarrier) Set(key, val string) {
-	if strings.Contains(key, "traceid") || strings.Contains(strings.ToLower(key), "traceid") {
+func (t *tagsCarrier) Set(key, val string) {
+	if strings.Contains(strings.ToLower(key), "traceid") {
```
devnev: extract `strings.ToLower(key)`

tracing/opentracing/id_extract.go
```diff
@@ -12,39 +9,55 @@ import (
...
-	TagTraceId = "trace.traceid"
-	TagSpanId  = "trace.spanid"
+	TagTraceId           = "trace.traceid"
+	TagSpanId            = "trace.spanid"
+	TagSampled           = "trace.sampled"
+	JaegerNotSampledFlag = "0"
 )
 
-// hackyInjectOpentracingIdsToTags writes the given context to the ctxtags.
+// injectOpentracingIdsToTags writes the given context to the ctxtags if the trace.
```
devnev: Fix comment

domgreen:
@devnev ping

tracing/opentracing/id_extract.go
```diff
@@ -12,39 +9,56 @@ import (
...
-	} else if (strings.Contains(key, "spanid") && !strings.Contains(key, "parent")) || (strings.Contains(strings.ToLower(key), "spanid") && !strings.Contains(strings.ToLower(key), "parent")) {
+	}
+
+	if strings.Contains(key, "spanid") && !strings.Contains(strings.ToLower(key), "parent") {
 		t.Tags.Set(TagSpanId, val) // this will most likely be base-16 (hex) encoded
-	} else if key == "uber-trace-id" {
+	}
+
+	if strings.Contains(key, "sampled") {
+		t.Tags.Set(TagSampled, val)
```
devnev: How about enforcing that the val is either `true` or `false`, and ignoring it if it isn't?

domgreen: We could do that however both OpenZipkin and OpenTracing do `strconv.FormatBool(sc.Sampled)` to set the value

domgreen: I think I would prefer to pass thru what is in the sampled field from the OT implementation if it is not true/false it would be valuable to have this visualised.

domgreen: @devnev is this change blocking a merge?

tracing/opentracing/id_extract.go
```diff
@@ -1,50 +1,69 @@
...
 		t.Tags.Set(TagSpanId, val) // this will most likely be base-16 (hex) encoded
-	} else if key == "uber-trace-id" {
+	}
+
+	if strings.Contains(key, "sampled") {
+		switch val {
+		case "true", "false":
+			t.Tags.Set(TagSampled, val)
+		default:
+			t.Tags.Set(TagSampled, "")
```
devnev: Why set to empty value instead of leaving unset?

domgreen: 😂 this was what you suggested during our conversation ... have updated :+1:

PR #191, opened 2019-01-30
Update Deps 

PrasadG193:
We are updating `golang.org/x/net` deps to release-branch.go1.11 in Kubernetes but we are facing a problem as go-grpc-middleware still uses the older version of `golang.org/x/net` package.

Will it be a good idea to update Gopkg.lock so that it uses the latest version of dependencies?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: false
 language: go
 go:
-  - 1.8.x
+  - 1.11.x
 env:
   - DEP_VERSION="0.3.2"
 
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -4,14 +4,14 @@
 [[projects]]
   name = "cloud.google.com/go"
   packages = ["compute/metadata"]
-  revision = "2d3a6656c17a60b0815b7e06ab0be04eacb6e613"
-  version = "v0.16.0"
+  revision = "28a4bc8c44b3acbcc482cff0cdf7de29a4688b61"
+  version = "v0.35.1"
 
 [[projects]]
   name = "github.com/davecgh/go-spew"
   packages = ["spew"]
-  revision = "346938d642f2ec3594ed81d874461961cd0faa76"
-  version = "v1.1.0"
+  revision = "8991bc29aa16c548c550c7ff78260e27b9ab7c73"
+  version = "v1.1.1"
 
 [[projects]]
   name = "github.com/gogo/protobuf"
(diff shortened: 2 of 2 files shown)


PrasadG193:
I signed it!

dims:
yes please!

wenjiaswe:
@PrasadG193 thanks for doing this!

@domgreen Would you please help to review this PR? I am not sure I am bothering the correct person I just looked up the file history:)

I just want to give a little more background on this. The dependency update we are doing in kubernetes is because there is a deadlock problem in the older version of golang/x/net, we are trying to get this dependency update fix in kubernetes v1.14, which is close soon. 

Xref: https://github.com/kubernetes/kubernetes/pull/71501#issuecomment-456286768

Thank you very much in advance!

wenjiaswe:
Hello @devnev ,I am not quite familiar with this community, I looked up some other PRs, please don't mind but would you mind please take a look at this change? We need to get this in so we could merge the kubernetes PR [kubernetes/kubernetes#71501 (comment)](https://github.com/kubernetes/kubernetes/pull/71501#issuecomment-456286768)

Thank you very much!

domgreen:
It looks from this PR we are updating the version of `go` that this project uses not just the package.

domgreen:
All tests pass with the update of deps so happy with that ... just wondering about the go version upgrede now.

.travis.yml
```diff
@@ -1,7 +1,7 @@
 sudo: false
 language: go
 go:
-  - 1.8.x
+  - 1.11.x
```
PrasadG193: @domgreen I had to update go version to make CI build happy. After updating dependencies with "dep ensure -update", `make all` was failing with `go 1.8`.

dims:
@PrasadG193 @domgreen i'd support the go version upgrade to help move to newer golang.org/x/net +1

.travis.yml
PrasadG193: Hey @domgreen, please let me know if we are fine with this change or you have any suggestions.

dims:
@domgreen @mwitkow This is holding up moving kubernetes to newer dependencies of docker etc. Can you please suggest a way forward with this?

The critical part is the movement of the httplex package:
https://github.com/grpc-ecosystem/go-grpc-middleware/pull/191/files#diff-bd247e83efc3c45ae9e8c47233249f18R84

dims:
related to https://github.com/kubernetes/kubernetes/issues/74413

wenjiaswe:
Thanks! @PrasadG193 @domgreen @dims

PR #200, opened 2019-03-12
Fix function comments based on best practices from Effective Go

CodeLingoTeam:
Hi, we updated an exported function comment based on best practices from [Effective Go](https://golang.org/doc/effective_go.html#commentary). It’s admittedly a relatively minor fix up. Does this help you?


domgreen:
@CodeLingoTeam thanks for the PR every little helps :D

PR #204, opened 2019-04-15
Fix span finish on client stream close (fix #196)

### LINKED ISSUE #196: tracedClientStream.CloseSend() doesn't finish the ClientSpan
This library's StreamClientInterceptor seems to work by wrapping streaming RPCs in a `tracedClientStream` that finishes the RPC's ClientSpan if anything goes wrong in the course of sending it. However, by my reading, there's no codepath that calls `s.finishClientStream` in the non-error case.

I think that `tracedClientStream.CloseSend()` may be the indended mechanism for finish the streaming RPC's ClientSpan, but it only calls `finishClientSpan` if calling `CloseSend()` on the wrapped stream returns an error: https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/tracing/opentracing/client_interceptors.go#L79

I believe this means that the ClientSpan is typically never closed.

inn0kenty:


inn0kenty:
I signed it!

PR #214, opened 2019-06-10
grpc_middleware: simplify ChainStreamClient

rogpeppe:
It's not necessary to use mutable closure variables to
keep track of which interceptor to call. Instead we
can use a more functional style, passing in the required
closure instead.

Using mutable closure variables is potentially unsafe if someone
decides to execute streamer functions concurrently (unlikely
but possible), is arguably harder to reason about, and
looks like a possible error when reviewing the code.

The code also ends up a bit smaller.

This style change could potentially be applied to other the other
interceptor chainers too.


rogpeppe:
Looking through the history, I find that the current code is very deliberate - my code uses O(chain-length) memory allocations and is somewhat slower (although faster than the original implementation as most of the closure allocations are outside the loop). Closing this accordingly.

PR #220, opened 2019-07-14
Refactor interceptor chain functions

jun06t:
Current chain logic is a bit redundant.
So I refactored it.

### DIFF
--- a/chain.go
+++ b/chain.go
@@ -18,35 +18,19 @@ import (
 func ChainUnaryServer(interceptors ...grpc.UnaryServerInterceptor) grpc.UnaryServerInterceptor {
 	n := len(interceptors)
 
-	if n > 1 {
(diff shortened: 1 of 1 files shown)


jun06t:
I signed it!

mbyio:
@domgreen @jun06t I think this is how they were originally implemented, but they were switched to a more efficient implementation here: https://github.com/grpc-ecosystem/go-grpc-middleware/pull/78. So this PR switched it back to the less efficient implementation.

Chaining middleware in this way is inherently inefficient because it forces the creation of closures for every request (as opposed to the way people chain HTTP middleware, with `type Middleware func(h http.Handler) http.Handler` which could be done with GRPC interceptors). Because it is inherently inefficient, I'm not sure if we should care. But I wanted to make sure you saw the possible regression.

PR #221, opened 2019-07-14
gRPC Zap ReplaceGrpcLoggerV2

kush-patel-hs:
We at Hootsuite were noticing noisy logs coming from grpc and we noticed it's because grpc_zap implements logger v1 interface. We need the logger v2 implemented instead, so I have opened this PR.

This is a revived and comments addressed version of #144 which was opened by @jamisonhyatt. Feel free to just merge this in after approval if it looks good!

### DIFF
--- a/logging/zap/examples_test.go
+++ b/logging/zap/examples_test.go
@@ -4,10 +4,10 @@ import (
 	"context"
 	"time"
 
-	"github.com/grpc-ecosystem/go-grpc-middleware"
-	"github.com/grpc-ecosystem/go-grpc-middleware/logging/zap"
-	"github.com/grpc-ecosystem/go-grpc-middleware/tags"
-	"github.com/grpc-ecosystem/go-grpc-middleware/tags/zap"
+	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
+	grpc_zap "github.com/grpc-ecosystem/go-grpc-middleware/logging/zap"
+	grpc_ctxtags "github.com/grpc-ecosystem/go-grpc-middleware/tags"
+	ctx_zap "github.com/grpc-ecosystem/go-grpc-middleware/tags/zap"
 	pb_testproto "github.com/grpc-ecosystem/go-grpc-middleware/testing/testproto"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
@@ -26,7 +26,7 @@ func Example_initialization() {
 		grpc_zap.WithLevels(customFunc),
 	}
 	// Make sure that log statements internal to gRPC library are logged using the zapLogger as well.
-	grpc_zap.ReplaceGrpcLogger(zapLogger)
+	grpc_zap.ReplaceGrpcLoggerV2(zapLogger)
 	// Create a server, make sure we put the grpc_ctxtags context before everything else.
 	_ = grpc.NewServer(
 		grpc_middleware.WithUnaryServerChain(
--- a/logging/zap/examples_test.go
+++ b/logging/zap/examples_test.go
@@ -26,7 +26,7 @@ func Example_initialization() {
 		grpc_zap.WithLevels(customFunc),
 	}
 	// Make sure that log statements internal to gRPC library are logged using the zapLogger as well.
-	grpc_zap.ReplaceGrpcLoggerV2(zapLogger, grpc_zap.DefaultGrpcLoggerV2Options)
+	grpc_zap.ReplaceGrpcLoggerV2(zapLogger)
 	// Create a server, make sure we put the grpc_ctxtags context before everything else.
 	_ = grpc.NewServer(
 		grpc_middleware.WithUnaryServerChain(
--- a/logging/zap/examples_test.go
+++ b/logging/zap/examples_test.go
@@ -26,7 +26,7 @@ func Example_initialization() {
 		grpc_zap.WithLevels(customFunc),
 	}
 	// Make sure that log statements internal to gRPC library are logged using the zapLogger as well.
-	grpc_zap.ReplaceGrpcLoggerV2(zapLogger)
+	grpc_zap.ReplaceGrpcLoggerV2(zapLogger, grpc_zap.DefaultGrpcLoggerV2Options)
 	// Create a server, make sure we put the grpc_ctxtags context before everything else.
 	_ = grpc.NewServer(
 		grpc_middleware.WithUnaryServerChain(
--- a/logging/zap/grpclogger.go
+++ b/logging/zap/grpclogger.go
@@ -65,12 +65,6 @@ func ReplaceGrpcLoggerV2(logger *zap.Logger, opts *GrpcLoggerV2Options) {
 	grpclog.SetLoggerV2(zgl)
 }
 
-const (
-	errorLevel = iota
-	warnLevel
-	infoLevel
-)
-
 type zapGrpcLoggerV2 struct {
 	*zap.Logger
 	verbosity int
--- a/logging/zap/grpclogger.go
+++ b/logging/zap/grpclogger.go
@@ -53,8 +53,11 @@ func ReplaceGrpcLoggerV2(logger *zap.Logger) {
 }
 
 // ReplaceGrpcLoggerV2WithVerosity replaces the grpc_.LoggerV2 with the provided logger and verbosity.
-func ReplaceGrpcLoggerV2WithVerosity(logger *zap.Logger, verosity int) {
-	zgl := &zapGrpcLoggerV2{Logger: logger.With(SystemField, zap.Bool("grpc_log", true))}
+func ReplaceGrpcLoggerV2WithVerosity(logger *zap.Logger, verbosity int) {
+	zgl := &zapGrpcLoggerV2{
+		Logger:    logger.With(SystemField, zap.Bool("grpc_log", true)),
+		verbosity: verbosity,
+	}
 	grpclog.SetLoggerV2(zgl)
 }
 
--- a/logging/zap/grpclogger.go
+++ b/logging/zap/grpclogger.go
@@ -5,8 +5,6 @@ package grpc_zap
 
 import (
 	"fmt"
-	"os"
-	"strconv"
 
 	"go.uber.org/zap"
 	"google.golang.org/grpc/grpclog"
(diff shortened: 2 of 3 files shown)


logging/zap/examples_test.go
```diff
@@ -4,10 +4,10 @@ import (
...
 	"time"
 
-	"github.com/grpc-ecosystem/go-grpc-middleware"
-	"github.com/grpc-ecosystem/go-grpc-middleware/logging/zap"
-	"github.com/grpc-ecosystem/go-grpc-middleware/tags"
-	"github.com/grpc-ecosystem/go-grpc-middleware/tags/zap"
+	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
+	grpc_zap "github.com/grpc-ecosystem/go-grpc-middleware/logging/zap"
+	grpc_ctxtags "github.com/grpc-ecosystem/go-grpc-middleware/tags"
+	ctx_zap "github.com/grpc-ecosystem/go-grpc-middleware/tags/zap"
```
kush-patel-hs: newer version of goimports does this because the package name is not the same as the last value in the path, let me know if you'll like me to undo this change

kush-patel-hs:
Other path would be instead of introducing ReplaceGrpcLoggerV2, just change ReplaceGrpcLogger internals. The interface to the user doesn't change (input is still only a zap.Logger).

kush-patel-hs:
I signed it!

kush-patel-hs:
@domgreen

kush-patel-hs:
@devnev

logging/zap/grpclogger.go
```diff
@@ -44,3 +47,103 @@ func (l *zapGrpcLogger) Printf(format string, args ...interface{}) {
 func (l *zapGrpcLogger) Println(args ...interface{}) {
 	l.logger.Info(fmt.Sprint(args...))
 }
+
+// ReplaceGrpcLoggerV2 replaces the grpc_log.LoggerV2 with the input zap.Logger
+// This logger adheres to the grpc go environment variables GRPC_GO_LOG_VERBOSITY_LEVEL and GRPC_GO_LOG_SEVERITY_LEVEL.
```
devnev: What is the use case for adding the extra log severity level handling instead of accepting the log level of the zap.Logger as-is?

kush-patel-hs: If I want my application to log at the info level and want the grpc system to log at error using the same global logger instance. Plus matches https://github.com/grpc/grpc-go/blob/24b2fb8959201be9ce659bc87b0d590a34c67eae/grpclog/loggerv2.go#L124. Thoughts? I can remove the severity level if you like and we could add it in the future if we feel the need.

devnev: I would favour starting with the explicit version: use the incoming logger's severity - and for verbosity, accept an argument with the desired verbosity level.

kush-patel-hs: What's your thoughts on having an option to allow for setting verbosity level and defaulting to 0?

kush-patel-hs: Updated @devnev

logging/zap/examples_test.go
domgreen: This needs to be fixed at some point not for this PR :+1:

logging/zap/grpclogger.go
```diff
@@ -44,3 +45,80 @@ func (l *zapGrpcLogger) Printf(format string, args ...interface{}) {
 func (l *zapGrpcLogger) Println(args ...interface{}) {
 	l.logger.Info(fmt.Sprint(args...))
 }
+
+// GrpcLoggerV2Options allows you to set grpclogger options when using ReplaceGrpcLoggerV2
+type GrpcLoggerV2Options struct {
+	verbosity int
+}
+
+// DefaultGrpcLoggerV2Options is a set of sane default options for ReplaceGrpcLoggerV2
```
domgreen: nit - would remove `sane` and add a `.` at the end of the sentence.

kush-patel-hs: i added `.` in all my exported struct/function comments :)

kush-patel-hs:
Thanks @domgreen! Pretty excited to get this in, will help us clear up some noisy logs from using the v1 logger at Hootsuite

logging/zap/grpclogger.go
```diff
@@ -44,3 +45,80 @@ func (l *zapGrpcLogger) Printf(format string, args ...interface{}) {
 func (l *zapGrpcLogger) Println(args ...interface{}) {
 	l.logger.Info(fmt.Sprint(args...))
 }
+
+// GrpcLoggerV2Options allows you to set grpclogger options when using ReplaceGrpcLoggerV2.
+type GrpcLoggerV2Options struct {
+	verbosity int
```
domgreen: This is not accessible :( you would never be able to set this.

logging/zap/grpclogger.go
```diff
@@ -44,3 +45,80 @@ func (l *zapGrpcLogger) Printf(format string, args ...interface{}) {
...
+}
+
+// DefaultGrpcLoggerV2Options is a set of default options for ReplaceGrpcLoggerV2.
+var DefaultGrpcLoggerV2Options = &GrpcLoggerV2Options{}
+
+// ReplaceGrpcLoggerV2 replaces the grpc_log.LoggerV2 with the input zap.Logger.
+// This logger allows you to set grpc verbosity level.
+func ReplaceGrpcLoggerV2(logger *zap.Logger, opts *GrpcLoggerV2Options) {
```
domgreen: Lets change this to `ReplaceLoggerV2(logger *zap.Logger)` this will match the upstream here https://github.com/grpc/grpc-go/blob/master/grpclog/loggerv2.go#L101

domgreen: Add a second `ReplaceLoggerV2WithVerbosity` with the option to pass in `v int` as https://github.com/grpc/grpc-go/blob/master/grpclog/loggerv2.go#L107 again matching upstream

domgreen: Change `ReplaceLoggerV2` to call `ReplaceLoggerV2WithVerbosity(logger , 0)` see https://github.com/grpc/grpc-go/blob/master/grpclog/loggerv2.go#L102

logging/zap/grpclogger.go
```diff
@@ -44,3 +45,80 @@ func (l *zapGrpcLogger) Printf(format string, args ...interface{}) {
...
 	l.logger.Info(fmt.Sprint(args...))
 }
+
+// GrpcLoggerV2Options allows you to set grpclogger options when using ReplaceGrpcLoggerV2.
+type GrpcLoggerV2Options struct {
+	verbosity int
+}
+
+// DefaultGrpcLoggerV2Options is a set of default options for ReplaceGrpcLoggerV2.
+var DefaultGrpcLoggerV2Options = &GrpcLoggerV2Options{}
```
domgreen: remove this and GrpcLoggerV2Options once done the below comments :+1:

logging/zap/grpclogger.go
```diff
@@ -44,3 +45,80 @@ func (l *zapGrpcLogger) Printf(format string, args ...interface{}) {
...
+
+// ReplaceGrpcLoggerV2 replaces the grpc_log.LoggerV2 with the input zap.Logger.
+// This logger allows you to set grpc verbosity level.
+func ReplaceGrpcLoggerV2(logger *zap.Logger, opts *GrpcLoggerV2Options) {
+	zgl := &zapGrpcLoggerV2{Logger: logger.With(SystemField, zap.Bool("grpc_log", true))}
```
domgreen: Inside Relace With Verbosity you can then do :
```
zgl := &zapGrpcLoggerV2{
  Logger: logger.With(SystemField, zap.Bool("grpc_log", true)),
  verbosity: 0,
}
```

logging/zap/grpclogger.go
kush-patel-hs: 🤔 I can do that! Want to drop the Grpc from the name?

logging/zap/grpclogger.go
kush-patel-hs: Oh whoops lol :(

logging/zap/grpclogger.go
kush-patel-hs: done

logging/zap/grpclogger.go
kush-patel-hs: done

logging/zap/grpclogger.go
```diff
@@ -44,3 +45,75 @@ func (l *zapGrpcLogger) Printf(format string, args ...interface{}) {
 func (l *zapGrpcLogger) Println(args ...interface{}) {
 	l.logger.Info(fmt.Sprint(args...))
 }
+
+// ReplaceGrpcLoggerV2 replaces the grpc_log.LoggerV2 with the input zap.Logger.
+// This logger allows you to set grpc verbosity level.
```
domgreen: This is only the WithVerbosity one

logging/zap/grpclogger.go
```diff
@@ -44,3 +45,75 @@ func (l *zapGrpcLogger) Printf(format string, args ...interface{}) {
...
+func ReplaceGrpcLoggerV2(logger *zap.Logger) {
+	ReplaceGrpcLoggerV2WithVerosity(logger, 0)
+}
+
+// ReplaceGrpcLoggerV2WithVerosity replaces the grpc_.LoggerV2 with the provided logger and verbosity.
+func ReplaceGrpcLoggerV2WithVerosity(logger *zap.Logger, verbosity int) {
```
domgreen: Verosity > Verbosity

kush-patel-hs: done

logging/zap/grpclogger.go
kush-patel-hs: true, done

kush-patel-hs:
Thanks for all the time you two spent on reviewing and providing feedback so far @domgreen @devnev :) I think I've addressed all feedbacks now, let's push this through :)

kush-patel-hs:
Thanks for all the help 👍

fproulx-eoscanada:
@kush-patel-hs Hi! I tried this. My default `zap.Logger` instance is set to `Info` level and it appears that setting the severity to `2` like `grpc_zap.ReplaceGrpcLoggerV2WithVerbosity(zlog, 2)` does not lower the noise. Skimming through the code this `severity` appears ? to be ignored ? So should I create a different `zap.Logger` with `Error` level before calling `ReplaceGrpcLoggerV2WithVerbosity` ?

kush-patel-hs:
@fproulx-eoscanada verbosity `2` will increase noise, the higher the verbosity number the more noise (i.e max noise is 99). Less noise is 0. Verbosity will be set to 0 if you simply do `grpc_zap.ReplaceGrpcLoggerV2(zlog)`. You would notice the LoopyWriter logs transport closing in particular go away changing from verbosity 2 to verbosity 0. If you want even less noise than that than yes create a new zap logger with Error or Warn level before doing the replace.

fproulx-eoscanada:
@kush-patel-hs ah! great so let's set it back to `0` and try again =)
Thanks -- great contribution by Hootsuite. We love grpc and zap too =)

kush-patel-hs:
Awesome :) let me know if you need any more help! 
And thank you, appreciate it :)

Icedroid:
Who can add logrus log with ReplaceGrpcLoggerV2?

srp:
In case others run across this thread, you might want or need to pass a sublogger with a different level then your root logger. This can look something like:

```
grpc_zap.ReplaceGrpcLoggerV2(logger.WithOptions(zap.IncreaseLevel(zap.WarnLevel))
```

PR #226, opened 2019-08-27
Move from dep to modules

domgreen:
Adding support for go modules.

The main concern here should be that we do not want to break backwards compatibility with people that are currently using the project. After thinking and discussing it with a few we think that this should be fine as it is. Anyone currently using the project should be okay as we do not have our vendor code checked in and they would be using their own dependencies in some manner (eg. dep).

cc @johanbrandhorst @bwplotka


domgreen:
I have tested this locally with an internal repo and results look good.

@devnev @mwitkow waiting on a tick from one of you guys :+1:

PR #234, opened 2019-09-20
adding v2 grpc logger

domgreen:
This brings it back to parity with zap

### DIFF
--- a/logging/logrus/grpclogger.go
+++ b/logging/logrus/grpclogger.go
@@ -4,6 +4,8 @@
 package grpc_logrus
 
 import (
+	"fmt"
+
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc/grpclog"
 )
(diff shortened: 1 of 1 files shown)


logging/logrus/grpclogger.go
```diff
@@ -4,12 +4,70 @@
...
 
 // ReplaceGrpcLogger sets the given logrus.Logger as a gRPC-level logger.
 // This should be called *before* any other initialization, preferably from init() functions.
 func ReplaceGrpcLogger(logger *logrus.Entry) {
-	grpclog.SetLogger(logger.WithField("system", SystemField))
+	grpclog.SetLoggerV2(&logrusGrpcLoggerV2{logger: logger.WithField("system", SystemField)})
+}
+
+type logrusGrpcLoggerV2 struct {
+	logger *logrus.Entry
```
johanbrandhorst: If we embed this instead we can just define the methods that don't already exist on `*logrus.Entry`, i.e., it removes the need to define `Info`,  `Infof`, `Error`, etc.

logging/logrus/grpclogger.go
```diff
@@ -4,12 +4,70 @@
...
+func (l *logrusGrpcLoggerV2) Info(args ...interface{}) {
+	l.logger.Info(args...)
+}
+
+func (l *logrusGrpcLoggerV2) Infoln(args ...interface{}) {
+	l.logger.Info(args...)
+}
+
+func (l *logrusGrpcLoggerV2) Infof(format string, args ...interface{}) {
+	l.logger.Info(fmt.Sprintf(format, args...))
```
johanbrandhorst: Why not `Infof`?

PR #240, opened 2019-10-19
Make SetInContext and NewTags to be public.

### LINKED ISSUE #182: make rpc_ctxtags.SetInContext  public access
for some reason, we want to copy `rpc_ctxtags.Tags` from one to another `context.Context`, and looks like it private function to do it,  is it  possible to make this method public?

mkorolyov:
In case if you create grpc client not in grpc service
you don't have possibility to set tags.
This patch makes some private API
to be public and fix described case.
fixes #182.

### DIFF
--- a/tags/context.go
+++ b/tags/context.go
@@ -69,10 +69,10 @@ func Extract(ctx context.Context) Tags {
 	return t
 }
 
-func setInContext(ctx context.Context, tags Tags) context.Context {
+func SetInContext(ctx context.Context, tags Tags) context.Context {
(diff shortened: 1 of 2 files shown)


mkorolyov:
@googlebot I signed it!

mkorolyov:
@googlebot I fixed it.

mkorolyov:
@googlebot I fixed it.

johanbrandhorst:
The commit still has two authors, might be easiest to create a new PR.

PR #247, opened 2019-11-23
recovery: add custom func on recovery example

Shikugawa:
It seems that handler function of recovery is missing on code example of recovery. It may cause confusion for users. But, there is valid handler function on test suite so add them.


Shikugawa:
@googlebot I signed it!

johanbrandhorst:
Thanks for your contribution!

PR #250, opened 2019-11-28
fix BackoffExponential's godoc.

syokomiz:
The Backoff Exponential function is wrong.
If the first attmpt is 100ms, the 5th attmpt should be 1.6s.

1th 100ms
2nd 200ms
3rd 400ms
4th 800ms
5th 1600ms = 1.6s


syokomiz:
@googlebot I signed it!

PR #257, opened 2020-01-04
Renew deprecated code

khasanovbi:


johanbrandhorst:
Could you please rebase this on master?

auth/auth_test.go
```diff
@@ -40,12 +41,17 @@ func buildDummyAuthFunction(expectedScheme string, expectedToken string) func(ct
...
 		}
 		return context.WithValue(ctx, authedMarker, "marker_exists"), nil
 	}
 }
 
+func getCodeFromError(err error) codes.Code {
+	st, _ := status.FromError(err)
+	return st.Code()
+}
+
```
johanbrandhorst: Can we use `status.Code` instead?

logging/common.go
```diff
@@ -8,16 +8,17 @@ import (
...
 )
 
 // ErrorToCode function determines the error code of an error
 // This makes using custom errors with grpc middleware easier
 type ErrorToCode func(err error) codes.Code
 
 func DefaultErrorToCode(err error) codes.Code {
-	return grpc.Code(err)
+	st, _ := status.FromError(err)
+	return st.Code()
```
johanbrandhorst: Can we use `status.Code` instead?

recovery/interceptors_test.go
```diff
@@ -16,13 +16,24 @@ import (
...
 	panicPing = &pb_testproto.PingRequest{Value: "panic", SleepTimeMs: 9999}
 )
 
+func getCodeFromError(err error) codes.Code {
+	st, _ := status.FromError(err)
+	return st.Code()
+}
+
+func getDescFromError(err error) string {
+	st, _ := status.FromError(err)
```
johanbrandhorst: How about `status.Convert(err).Message()`?

khasanovbi:
Oh, I understand, I have dep based vendors with old grpc requirement, that doesn't have Convert, Code functions  https://github.com/grpc-ecosystem/go-grpc-middleware/blob/f849b5445de4819127e123ca96ba0eeb62b5e479/Gopkg.lock#L116. 
Now I see that repo migrated to modules.

johanbrandhorst:
Thanks for your contribution!

PR #260, opened 2020-02-03
Reset connect backoff

iostrovok:
support https://godoc.org/google.golang.org/grpc#ClientConn.ResetConnectBackoff


johanbrandhorst:
Hi @iostrovok. I think we should discuss whether we want to add this before going straight to implementing it. Reading the documentation for the method, I'm not convinced:

> In general, this function should not be used. Typical service or network outages result in a reasonable client reconnection strategy by default.

(https://godoc.org/google.golang.org/grpc#ClientConn.ResetConnectBackoff)

What is your use case that requires adding this?

yashrsharma44:
Hi @iostrovok, any updates to this PR?

PR #262, opened 2020-02-06
feature(tracing): Add datadog support to injectOpenTracingIdsToTags

wdullaer:
Datadog uses a convention for it's keys that is currently not supported.

See https://github.com/DataDog/dd-trace-go/blob/v1/ddtrace/tracer/textmap.go#L77 and https://github.com/DataDog/dd-trace-go/blob/v1/ddtrace/tracer/textmap.go#L221 for some background.

This PR adds a few additional heuristics to `injectOpenTracingIdsToTags` which should help capture these.

Alternatively, if you are ok with using reflection: you can cast the span to a ddtrace.SpanContext which does give direct access to the spanid and traceid, allowing you to bypass the inject call altogether.


wdullaer:
@googlebot I signed it8

PR #263, opened 2020-02-06
pass FullMethod to AuthFunc

orangetangerine:
Sometime we need to call methods without authentication, like /package.service/healthz and /Stat or even /Auth. We need a method name filter in this case.
Although it can be done by implementing AuthFuncOverride, or make another Interceptor,
it would be nice that these things can be done in one package.

Since it's just a small change, I think there is no need to do backward compatibility.


johanbrandhorst:
It should be possible to get the method from the context already: https://godoc.org/google.golang.org/grpc#Method. Does this cover your use case?

Unfortunately, however small the change, we absolutely _cannot_ break compatibility like this, so if you need something more, please open an issue and we can discuss solutions.

PR #265, opened 2020-02-10
Add unary client interceptor for validation

sashayakovtseva:
Sometimes it is necessary to validate grpc request before actually sending it. 
With `UnaryClientInterceptor` one can do so.


sashayakovtseva:
@googlebot I signed it!

PR #268, opened 2020-02-20
add build target for go 1.13 for travis.

panshul007:
This will add a build target or go version 1.13 for travis.


panshul007:
@googlebot I signed it!

PR #272, opened 2020-02-24
pass context to logrus entry

olomix:
When you use custom logrus formatter, ctx may be used to pass additional information for formatter (for example x-request-id to include into log records). I've made server interceptor to pass this context to logrus entry.


olomix:
@googlebot I signed it!

olomix:
@googlebot I signed it!

PR #281, opened 2020-03-29
Propery handle nil panics

misberner:
Checking that `recover()` returns a non-`nil` value is not sufficient to determine whether a panic occurred (see https://github.com/golang/go/issues/25448 for a discussion and possible workaround). This PR implements the proper logic to determine if a panic occurs and invokes the recovery handler (either the default or custom one) with the actual panic value, even if it was `nil`.

See the first commit in this PR (`6f8c126`) for a test case that exhibits the problem on the existing code.


misberner:
@googlebot I signed it!

johanbrandhorst:
Seems googlebot isn't happy, can you try again?

misberner:
@googlebot I signed it!

misberner:
> Seems googlebot isn't happy, can you try again?

Done. Looks like my employer hadn't signed the CLA yet earlier

johanbrandhorst:
Thanks for your contribution!

PR #283, opened 2020-04-03
Update doc.go to add missing parens on the example

scottcrawford03:
When reading the docs I noticed that there was a missing paren on the example. This does the heavy lifting to add that paren back in 💃


johanbrandhorst:
Thanks for your contribution!

PR #284, opened 2020-04-06
Log message producer

### LINKED ISSUE #258: Option to use original error message instead of generic interceptor messages
The ["finished unary call with code" message](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/3c51f7f332123e8be5a157c0802a228ac85bf9db/logging/zap/server_interceptors.go#L66) is not very helpful as main error message in error logs.
The actual error message is often more important.

Services should have the option to use the original error message as main `message` (in addition to adding structured `error` and `code` fields). The interceptor's own message if required at all could be added as `interception` context in this case.

This will make features that work mainly on the error message, such as [Datadog's pattern detection](https://docs.datadoghq.com/logs/explorer/patterns/) much more useful, without losing the benefits of structured logging.

tinogoehlert:
Adds `WithMessageProducer` functions for log formation.

resolves #258

### DIFF
--- a/logging/logrus/client_interceptors_test.go
+++ b/logging/logrus/client_interceptors_test.go
@@ -2,8 +2,6 @@ package grpc_logrus_test
 
 import (
 	"io"
-	"runtime"
-	"strings"
 	"testing"
 
 	grpc_logrus "github.com/grpc-ecosystem/go-grpc-middleware/logging/logrus"
@@ -27,10 +25,6 @@ func customClientCodeToLevel(c codes.Code) logrus.Level {
 }
 
 func TestLogrusClientSuite(t *testing.T) {
-	if strings.HasPrefix(runtime.Version(), "go1.7") {
-		t.Skipf("Skipping due to json.RawMessage incompatibility with go1.7")
-		return
-	}
 	opts := []grpc_logrus.Option{
 		grpc_logrus.WithLevels(customClientCodeToLevel),
 	}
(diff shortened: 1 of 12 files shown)


tinogoehlert:
@googlebot I signed it!

logging/zap/client_interceptors.go
```diff
@@ -8,6 +8,8 @@ import (
 	"path"
 	"time"
 
+	"github.com/grpc-ecosystem/go-grpc-middleware/logging/zap/ctxzap"
```
johanbrandhorst, line 11: Combine this with the 2nd import group

logging/logrus/client_interceptors_test.go
```diff
@@ -187,3 +187,40 @@ func (s *logrusClientOverrideSuite) TestPingList_HasOverrides() {
 	assert.NotContains(s.T(), msgs[0], "grpc.time_ms", "message must not contain default duration")
 	assert.Contains(s.T(), msgs[0], "grpc.duration", "message must contain overridden duration")
 }
+
+func TestZapLoggingClientMessageProducerSuite(t *testing.T) {
+	if strings.HasPrefix(runtime.Version(), "go1.7") {
+		t.Skip("Skipping due to json.RawMessage incompatibility with go1.7")
```
johanbrandhorst, line 193: I wonder how long it will take for us to get to `go1.70.0` :thinking:

johanbrandhorst: Lets remove this, we don't support `Go 1.7`

logging/zap/shared_test.go
```diff
@@ -104,3 +105,12 @@ func (s *zapBaseSuite) getOutputJSONs() []map[string]interface{} {
 
 	return ret
 }
+
+func StubtMessageProducer(ctx context.Context, msg string, level zapcore.Level, code codes.Code, err error, duration zapcore.Field) {
```
johanbrandhorst, line 109:
```suggestion
func StubMessageProducer(ctx context.Context, msg string, level zapcore.Level, code codes.Code, err error, duration zapcore.Field) {
```

logging/logrus/client_interceptors_test.go
tinogoehlert: okay 😄 

i will remove it in the files i've touched. The PR would get messy otherwise ^^

johanbrandhorst:
Thanks for your contribution!

PR #290, opened 2020-05-01
Moved imports to v2; Moved to Go 1.14.2

bwplotka:
Following https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher we need this. It might be still non-buildable.

Signed-off-by: Bartlomiej Plotka <bwplotka@gmail.com>


go.mod
```diff
@@ -11,4 +11,4 @@ require (
 	google.golang.org/grpc v1.19.0
 )
 
-go 1.13
+go 1.14
```
johanbrandhorst, line 14: Nit: can we move this above the `require` statement?

interceptors/logging/interceptors_test.go
```diff
@@ -9,18 +9,18 @@ import (
...
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/logging"
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/tags"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 
-	pb_testproto "github.com/grpc-ecosystem/go-grpc-middleware/grpctesting/testproto"
+	pb_testproto "github.com/grpc-ecosystem/go-grpc-middleware/v2/grpctesting/testproto"
```
johanbrandhorst, line 23: What import grouping do we want to use? This is a little inconsistent. I usually do 3 import groups. Maybe now is a good time to change?

interceptors/server_test.go
```diff
@@ -11,10 +11,10 @@ import (
 	"testing"
 	"time"
 
-	"github.com/grpc-ecosystem/go-grpc-middleware/grpctesting"
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/grpctesting"
 	"google.golang.org/grpc/status"
 
-	pb_testproto "github.com/grpc-ecosystem/go-grpc-middleware/grpctesting/testproto"
+	pb_testproto "github.com/grpc-ecosystem/go-grpc-middleware/v2/grpctesting/testproto"
```
johanbrandhorst, line 17: :scream:

interceptors/logging/interceptors_test.go
bwplotka: so I use goimports which does 2 groups. What do you prefer? I am fine with whatever as long as there is popular auto formatting tool for this (:

interceptors/server_test.go
bwplotka: yea not formatted well

bwplotka:
Ok I moved to Go v2 , but let's fix formatting later, go list does not work without tag now

go.mod
bwplotka: done

interceptors/server_test.go
bwplotka: will fix in later PR

Issue #275, opened 2020-02-29
go-grpc-middleware.v2: Action Plan.


bwplotka:
Hi :wave: 

We have tons of stuff to improve in this library and there wasn't really time for this. Let's do it finally! Let's maintain this issue as a single place to track our work. :muscle: cc @domgreen @johanbrandhorst @mwitkow 

The plan is to start from scratch in the `v2` branch. We will work in the background until it's done.

## Work Done

* [x] Started v2 branch, implemented all while leaving same API.  Assignee: @bwplotka 
* [x] Never again use those weird `module_names` :upside_down_face: [Done](https://github.com/grpc-ecosystem/go-grpc-middleware/pull/276)  Assignee: @bwplotka 
* [X] Multi-module architecture, allowing us to avoid dependency hell, but also retaining mono repo. Module structure: [Done](https://github.com/grpc-ecosystem/go-grpc-middleware/pull/276)  Assignee: @bwplotka 
    The key part is to make sure the core is never importing the provider's code. There has to be indirection. This way the lib has minimum dependencies possible, with core literally having core only. 

   If maintaining multi modules will be problematic, we can look into [modularise](https://github.com/Helcaraxan/modularise-example-core) cc @Helcaraxan

### Blockers:

* [ ] Do one more review of all interfaces we expose to check if we can simplify something. Assignee @johanbrandhorst 
* [ ] Upgrade gRPC and use https://godoc.org/google.golang.org/grpc#ChainUnaryInterceptor  Assignee: @bwplotka 
* [ ]  Upgrade golang/protobuf to v2. Same with gogo?  Assignee: @bwplotka
* [ ] Move to GitHub actions Assignee: @bwplotka 
* [ ] MIgration docs! 
* [ ] Start using `gorelease` (https://godoc.org/golang.org/x/exp/cmd/gorelease)
* [ ] Consider adding whitelist behaviour to auth interceptor: How to enable whitelist auth?
* [ ] Add tracing interface and move opentracing to provides. Add opentelemetry providers.
* [ ] Use the v2 in [Thanos](https://thanos.io) for first hand expierience in upgrading. Assignee: @bwplotka  

### Can be done in 2.x

* [ ] Move https://github.com/grpc-ecosystem/go-grpc-prometheus as another provider. Deprecate repo.
* [ ] Add [failin](https://github.com/thanos-io/thanos/blob/master/Makefile#L305)t to avoid not needed imports (e.g v1 packages)
[...]

PR #297, opened 2020-06-03
Add zerolog support to logging

irridia:
Integrates zerolog into go-grpc-middleware/logging.

This is an adaptation of the logrus code but with the Deprecated pieces (that I noticed) removed, and modified for zerolog's mechanisms.

Right now, the imports reference the fork, but would of course be updated before merge. If preferred, I can add ctxzerolog first, but the testing imports also complicate matters. Tests pass, but `make all` is confused because of the imports.

I welcome all feedback!


irridia:
@googlebot I signed it!

bwplotka:
Amazing!

We love this work, but the idea is to develop on v2. Can you rebase on `v2` branch? We actually redesigned to it should be thousands times easier to integrate new logger :heart: 

What do you think?

You can read more here https://github.com/grpc-ecosystem/go-grpc-middleware/issues/275

nashikb:
> We love this work, but the idea is to develop on v2. Can you rebase on v2 branch? We actually redesigned to it should be thousands times easier to integrate new logger ❤️

Sure, will do. After poking at this code, several of the v2 tasks are <3 to me. :)

nashikb:
What do you think about adding this implementation to v1 *AND* I'll adjust/implement for v2? Seems like people will get the benefit sooner from v1.

irridia:
> We love this work, but the idea is to develop on v2. Can you rebase on v2 branch? We actually redesigned to it should be thousands times easier to integrate new logger ❤️

Sure, will do. After poking at this code, several of the v2 tasks are <3 to me. :)

irridia:
What do you think about adding this implementation to v1 AND I'll adjust/implement for v2? Seems like people will get the benefit sooner from v1.

irridia:
Done: https://github.com/grpc-ecosystem/go-grpc-middleware/pull/299

bwplotka:
> What do you think about adding this implementation to v1 AND I'll adjust/implement for v2? Seems like people will get the benefit sooner from v1.

I think we don't want to focus on v1. It's broken loop to add another 2 thousands lines of code to the library we want only support for bugfixes (: Also you had thousands benefits from switching to new version of go-grpc, so hopefully the effect of supporting zerolog only on `v2` will give even more motivation for initial pain of migrating (:

irridia:
> Also you had thousands benefits from switching to new version of go-grpc, so hopefully the effect of supporting zerolog only on `v2` will give even more motivation for initial pain of migrating (:

Is v2 at the point where it's a fully-functional replacement for v1? I agree if anyone can switch to v2 now.

bwplotka:
Let's close this one as v2 was merged: https://github.com/grpc-ecosystem/go-grpc-middleware/pull/299

PR #303, opened 2020-06-08
Update github.com/golang/protobuf to google.golang.org/protobuf

yashrsharma44:
This PR addresses the change for upgrading `github.com/golang/protobuf` to `google.golang.org/protobuf`

There were some issues while I was running `make proto`. I was getting an unused `_ google/protobuf` in `go-grpc-middleware/grpctesting/gogotestpb/fields.pb.go` so I manually deleted it from, however that is not supposed to be edited.

Another issue is failing tests, which I am not able to comprehend, so I would like some suggestions from the maintainers


go.mod
```diff
@@ -4,11 +4,11 @@ go 1.14
 
 require (
 	github.com/gogo/protobuf v1.2.1
-	github.com/golang/protobuf v1.3.2
+	github.com/golang/protobuf v1.4.1
```
yashrsharma44, line 7: While running `make test`, this line is regenerated again, so any pointers how to remove a module from the package itself?

johanbrandhorst: This won't disappear, the new library imports the old one still. This is fine.

interceptors/logging/payload.go
```diff
@@ -136,6 +135,7 @@ func logProtoMessageAsJson(logger Logger, pbMsg interface{}, key string, msg str
 		payload, err := (&jsonpbObjectMarshaler{pb: p}).marshalJSON()
 		if err != nil {
 			logger = logger.With(key, err.Error())
+			logger.Log(ERROR, err.Error())
```
johanbrandhorst, line 138: This change seems unrelated, is it a bug fix?

interceptors/logging/payload.go
```diff
@@ -17,7 +16,7 @@ import (
 var (
 	// JsonPbMarshaller is the marshaller used for serializing protobuf messages.
 	// If needed, this variable can be reassigned with a different marshaller with the same Marshal() signature.
-	JsonPbMarshaller JsonPbMarshaler = &jsonpb.Marshaler{}
+	JsonPbMarshaller JsonPbMarshaler = &jsonpb.MarshalOptions{}
```
johanbrandhorst, line 19: Would be nice to remove this global and make it a parameter to logging options or something. Could be another PR.

interceptors/logging/payload.go
```diff
@@ -148,9 +148,9 @@ type jsonpbObjectMarshaler struct {
...
 func (j *jsonpbObjectMarshaler) marshalJSON() ([]byte, error) {
-	b := &bytes.Buffer{}
-	if err := JsonPbMarshaller.Marshal(b, j.pb); err != nil {
+	b, err := JsonPbMarshaller.Marshal(j.pb)
+	if err != nil {
 		return nil, fmt.Errorf("jsonpb serializer failed: %v", err)
 	}
-	return b.Bytes(), nil
+	return b, nil
 }
```
johanbrandhorst, line 156: Can we replace uses of this function altogether with `JsonPbMarshaller.Marshal(pb)`?

interceptors/logging/payload_test.go
```diff
@@ -60,7 +60,6 @@ func TestPayloadSuite(t *testing.T) {
 func (s *loggingPayloadSuite) TestPing_LogsBothRequestAndResponse() {
 	_, err := s.Client.Ping(s.SimpleCtx(), goodPing)
 	require.NoError(s.T(), err, "there must be not be an error on a successful call")
-
```
johanbrandhorst, line 63: Was this intentional? I'd prefer to keep this newline

interceptors/logging/payload.go
yashrsharma44: Sure, will make an issue for it.

interceptors/logging/payload.go
yashrsharma44: Ohh my bad, I was debugging the failing tests and added this line. Will remove this.

interceptors/logging/payload_test.go
yashrsharma44: Sure

interceptors/logging/payload.go
yashrsharma44: Sure.

yashrsharma44: Shall I remove `jsonpbObjectMarshaler` from line [145](https://github.com/yashrsharma44/go-grpc-middleware/blob/update-proto/interceptors/logging/payload.go#L145)?

johanbrandhorst:
For the test cases, I'm guessing the race condition is because we're copying a protobuf value somewhere we should be using a reference instead. Not sure about the `mockedReporter` equality errors, will need to investigate more.

bwplotka:
Taking this over as agreed offline, will create new PR. Thanks @yashrsharma44 for good start. I will try to upgrade gRPC and gogo as well

PR #304, opened 2020-06-13
Retry dial and connection errors for grpc stream.

kartlee:
The change submitted is about retrying the grpc stream for dial and connection errors. With go 1.14  + google.golang.org/grpc (v1.29.1 tag), the observation is when the server is restarted,  the grpc stream fails to re-establish due to connection errors. The patch proposed checks if the error returned is a retriable one and uses the same backoff logic configured for the retry.

Let me know your comments on this.

-Karthik


kartlee:
@googlebot I signed it!

johanbrandhorst:
This looks great, thanks! Could you cherry pick this to v2 please?

kartlee:
Thanks for considering the change. I just raised one for v2 as you asked.

-Karthik

PR #307, opened 2020-06-15
Modify the require_clean_work_tree function

yashrsharma44:
Modify the require_clean_work_tree function. Taken from - https://github.com/git/git/blob/master/git-sh-setup.sh#L211


bwplotka:
This makes sense, but can we push it to `master`? Also worth to fix on Thanos as well??? (: Looks cleaner

yashrsharma44:
Sure, I will make a pr to Thanos.

bwplotka:
@googlebot I consent.

PR #311, opened 2020-06-29
Request Logging

### LINKED ISSUE #310: v2: Feature Request : Request/Response Logging
I am currently working on adding a logging feature of grpc, at the request/response level for [Thanos](https://github.com/thanos-io/thanos/pull/2694). I had a look at the [current logging](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/v2/interceptors/logging/interceptors.go) implementation, and I found that the we don't log at a level where we could access the req/resp object.

[Payload](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/v2/interceptors/logging/payload.go) implementation has the feature that I am looking for, but I want similar stuff _minus_ payload logging.

More technically, the current logging does not implement the [`PostMsgSend`](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/v2/interceptors/logging/interceptors.go#L53) and [`PostMsgRecv`](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/v2/interceptors/logging/interceptors.go#L55) method, so I wanted to extend the reporter by adding implementation logic for `PostMsgSend` and `PostMsgRecv`.
* I guess having a new reporter would be more beneficial for not breaking backwards compatibility for the current logging implementation.
* Also, the [WithDecider](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/v2/interceptors/logging/options.go#L51) method only filters the logging based on error and method/service name, while I also wanted to filter out the logging based on the req/resp object.

I was willing to work on such implementation, so I wanted some suggestions / green flag, if adding this would be acceptable.

Ping @johanbrandhorst

yashrsharma44:
This PR achieves to add in a Request/Response logger in `grpc_middlewares` `v2` branch.

Fixes #310 

Current TODO -
- [x] ~~Initial setup for the RequestReporter~~ Configure the request logger in the old logger itself
- [x] Write tests for the new reporter
- [x] Write examples for the new reporter
- [x] **Decide what to log after we can access the request/response object** cc @bwplotka 
ref: https://github.com/grpc-ecosystem/go-grpc-middleware/pull/311/files#diff-0c98eb4ff4e54e07ef0a999e5d5bd678R137
- [x] Naming of the reporter - Is the current naming correct, or do we want something else :stuck_out_tongue: 

If any suggestions are there, feel free to add in more :sweat_smile: 


interceptors/logging/logging.go
```diff
@@ -55,6 +55,14 @@ func DefaultDeciderMethod(_ string, _ error) bool {
 	return true
 }
 
+// ClientRequestLoggingDecider is a user provided function for deciding whether to log the client-side
+// request/response metadata
+type ClientRequestLoggingDecider func(ctx context.Context, fullMethodName string, servingObject interface{}) bool
+
+// ServerRequestLoggingDecider is a user provided function for deciding whether to log the server-side
+// request/response metadata
+type ServerRequestLoggingDecider func(ctx context.Context, fullMethodName string, servingObject interface{}) bool
```
bwplotka, line 64: This would work, just curious if we don't want to compose those with decider below. For example would return `payload bool, request bool` or something

interceptors/logging/request_logging.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/tags"
+	"google.golang.org/grpc"
+)
+
+type serverRequestReporter struct {
+	ctx    context.Context
+	logger Logger
+}
+
+// Log the server/method name details and error if any
```
bwplotka, line 17: Comment in Go should start with Method name (`PostCall`) and finish with trailing period.

bwplotka: and all below

bwplotka: and all below

yashrsharma44: Sure,will take care of that

interceptors/logging/interceptors.go
```diff
@@ -26,12 +27,15 @@ type reporter struct {
 	typ             interceptors.GRPCType
 	service, method string
 
+	requestLoggerDecider bool
```
kakkoyun, line 30: I'd rather name this to 
```suggestion
	requestLoggerDecision bool
```

interceptors/logging/interceptors.go
```diff
@@ -50,64 +54,112 @@ func (c *reporter) PostCall(err error, duration time.Duration) {
...
 
-func (c *reporter) PostMsgReceive(interface{}, error, time.Duration) {}
+	// If the Request Logging is configured to be skipped.
+	if !c.requestLoggerDecider {
+		return
+	}
+	code := c.opts.codeFunc(err)
+	logger := c.logger.With(extractFields(tags.Extract(c.ctx))...)
+	logRequestResponse(c.ctx, logger.With("grpc.recv.duration", duration.String()), req, c.opts.levelFunc(code))
+
```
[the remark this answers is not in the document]
yashrsharma44: I thought that might be taken care of the formatter, but going to change it :)

interceptors/logging/interceptors.go
```diff
@@ -50,64 +54,105 @@ func (c *reporter) PostCall(err error, duration time.Duration) {
 	logger.With(c.opts.durationFieldFunc(duration)...).Log(c.opts.levelFunc(code), fmt.Sprintf("finished %s %s call", c.kind, c.typ))
 }
 
-func (c *reporter) PostMsgSend(interface{}, error, time.Duration) {}
+// PostMsgSend logs the details of the servingObject(req/resp) that is flowing out of the rpc.
+func (c *reporter) PostMsgSend(req interface{}, err error, duration time.Duration) {
```
bwplotka, line 58: This is not that easy. The problem is streaming. You got messages on every stream frame. This is not what we want probably

interceptors/logging/interceptors.go
```diff
@@ -50,64 +54,105 @@ func (c *reporter) PostCall(err error, duration time.Duration) {
...
-func StreamServerInterceptor(logger Logger, opts ...Option) grpc.StreamServerInterceptor {
+func StreamServerInterceptor(logger Logger, decider PostRequestLoggingDecider, opts ...Option) grpc.StreamServerInterceptor {
 	o := evaluateServerOpt(opts)
-	return interceptors.StreamServerInterceptor(&reportable{logger: logger, opts: o})
+	return interceptors.StreamServerInterceptor(&reportable{logger: logger, requestLoggerDecision: decider, opts: o})
+}
+
+// TODO: yashrsharma44
+// What are the things that we want to log?
+func logRequestResponse(ctx context.Context, logger Logger, servObj interface{}, level Level) {
```
bwplotka, line 143: Is it really response?

bwplotka:
There is easy way how to do what we want btw (:

interceptors/logging/interceptors.go
yashrsharma44: Sounds fair, maybe we need to check when does the stream start and end, so that we can the log rather than streaming all the frames.
Not sure how we would check the start and end of the streaming, need to dig into the grpc docs more.

interceptors/logging/interceptors.go
```diff
@@ -50,64 +54,105 @@ func (c *reporter) PostCall(err error, duration time.Duration) {
 	logger.With(c.opts.durationFieldFunc(duration)...).Log(c.opts.levelFunc(code), fmt.Sprintf("finished %s %s call", c.kind, c.typ))
 }
 
-func (c *reporter) PostMsgSend(interface{}, error, time.Duration) {}
+// PostMsgSend logs the details of the servingObject(req/resp) that is flowing out of the rpc.
+func (c *reporter) PostMsgSend(req interface{}, err error, duration time.Duration) {
+	// If the Request Logging is configured to be skipped.
```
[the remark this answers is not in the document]
yashrsharma44: Sure.

interceptors/logging/interceptors.go
yashrsharma44: From the client respect, the logging works the same manner, so it logs the response of the client.

yashrsharma44:
After having a 1:2 call with @bwplotka and @kakkoyun, we decided to modify the objectives of this PR - 
- [ ] Having a precall setup before the actual logging starts
* Decider to check if logging should be enabled (Implemented at the interceptor level. If not enabled, we return a `NoopReporter`)
* Decider to check if logging should be enabled for a given method( To be implemented using the options of the `grpc.Options`)
* Log the starting messages (We check using a boolean if the current message is the first and log it in the `PostMsgSend/PostMsgRecv` level)
- [ ] `PostMsgSend / PostMsgRecv` can log the payload (currently to be worked on in the next PR)
- [ ] `PostCall` to log the errors and corresponding error-logs 
#
This is a bit different from the original implementation because in the earlier PR, we were logging all the messages of the stream, which is quite verbose and irrelevant.

interceptors/logging/interceptors.go
```diff
@@ -50,64 +52,119 @@ func (c *reporter) PostCall(err error, duration time.Duration) {
...
+	c.firstMessageLogged = true
 
-func (c *reporter) PostMsgReceive(interface{}, error, time.Duration) {}
+	logger := c.logger.With(extractFields(tags.Extract(c.ctx))...)
+	logger = logger.With("grpc.recv.duration", duration.String())
+	if err != nil {
+		logger = logger.With("error", fmt.Sprintf("%v", err))
+	}
+
+	if c.isServer {
```
bwplotka, line 73: hm I think for a "start", for server it's recv always, and for client it's send, so maybe we can put `isServer` above and return immediately if server?

interceptors/logging/interceptors.go
```diff
@@ -50,64 +52,119 @@ func (c *reporter) PostCall(err error, duration time.Duration) {
...
+	c.firstMessageLogged = true
+
+	logger := c.logger.With(extractFields(tags.Extract(c.ctx))...)
+	logger = logger.With("grpc.recv.duration", duration.String())
+	if err != nil {
+		logger = logger.With("error", fmt.Sprintf("%v", err))
+	}
+
+	if c.isServer {
+		logger.Log("logged the start of the request.", fmt.Sprintf("request object: %v", req))
```
bwplotka, line 99: same here

interceptors/logging/logging.go
```diff
@@ -55,6 +55,10 @@ func DefaultDeciderMethod(_ string, _ error) bool {
 	return true
 }
 
+// PostRequestLoggingDecider is a user provided function for deciding whether the Request Logging should be configured.
+// Request Logging is enabled if the value of the function evaluates to true.
+type PostRequestLoggingDecider func(ctx context.Context, fullMethodName string, servingObject interface{}) bool
```
bwplotka, line 60: Do you mean PreRequest? Also "Request logging" is whole thing. I think you mean "start log entry", right?

yashrsharma44: Yeah, I didn't  change the naming after we had the discussion, will change it to PreRequest!

interceptors/logging/interceptors.go
```diff
@@ -50,64 +52,117 @@ func (c *reporter) PostCall(err error, duration time.Duration) {
...
 
-func (c *reporter) PostMsgReceive(interface{}, error, time.Duration) {}
+	logger := c.logger.With(extractFields(tags.Extract(c.ctx))...)
+	logger = logger.With("grpc.recv.duration", duration.String())
+	if err != nil {
+		logger = logger.With("error", fmt.Sprintf("%v", err))
+	}
+
+	logger.Log("logged the start of the request.", fmt.Sprintf("request object: %v", resp))
+
```
bwplotka, line 76: We do same thing twice, can we move to function?

interceptors/logging/interceptors.go
```diff
@@ -50,64 +52,117 @@ func (c *reporter) PostCall(err error, duration time.Duration) {
...
+	c.firstMessageLogged = true
 
-func (c *reporter) PostMsgReceive(interface{}, error, time.Duration) {}
+	logger := c.logger.With(extractFields(tags.Extract(c.ctx))...)
+	logger = logger.With("grpc.recv.duration", duration.String())
+	if err != nil {
+		logger = logger.With("error", fmt.Sprintf("%v", err))
+	}
+
+	logger.Log("logged the start of the request.", fmt.Sprintf("request object: %v", resp))
```
bwplotka, line 75:
```suggestion
	logger.Log("requested started.", fmt.Sprintf("request object: %v", resp))
```

interceptors/logging/interceptors_test.go
```diff
@@ -153,6 +154,12 @@ type loggingClientServerSuite struct {
 	*baseLoggingSuite
 }
 
+// customRequestLoggingDecider is a function which configures if request logging should be enabled.
```
bwplotka, line 157: No need for this comment I guess

interceptors/logging/interceptors.go
```diff
@@ -50,64 +52,117 @@ func (c *reporter) PostCall(err error, duration time.Duration) {
 	logger.With(c.opts.durationFieldFunc(duration)...).Log(c.opts.levelFunc(code), fmt.Sprintf("finished %s %s call", c.kind, c.typ))
 }
 
-func (c *reporter) PostMsgSend(interface{}, error, time.Duration) {}
+// PostMsgSend logs the details of the servingObject that is flowing out of the rpc.
+// resp object wrt server.
+// Log the details of the request, skip if the object is response.
```
bwplotka, line 57:
```suggestion
// Log the details of the first request, skip if the object is response.
``` 
? Same on `PostMsgRecv`

interceptors/logging/interceptors.go
yashrsharma44: I have made modifications, so this suggestion is outdated :stuck_out_tongue:

yashrsharma44:
Hey Everyone, this PR is ready to be merged.
Would love a final round of review - @kakkoyun @bwplotka

interceptors/logging/interceptors.go
```diff
@@ -47,40 +50,102 @@ func (c *reporter) PostCall(err error, duration time.Duration) {
 	if err != nil {
 		logger = logger.With("error", fmt.Sprintf("%v", err))
 	}
-	logger.With(c.opts.durationFieldFunc(duration)...).Log(c.opts.levelFunc(code), fmt.Sprintf("finished %s %s call", c.kind, c.typ))
+
+	if !c.opts.shouldLogRequest(interceptors.FullMethod(c.service, c.method), err) {
+		logger.With(c.opts.durationFieldFunc(duration)...).Log(c.opts.levelFunc(code), fmt.Sprintf("finished %s %s call", c.kind, c.typ))
```
bwplotka, line 55: we could just do it once I think :hugs: 

`logger.With(c.opts.durationFieldFunc(duration)...)`

interceptors/logging/interceptors.go
```diff
@@ -47,40 +50,102 @@ func (c *reporter) PostCall(err error, duration time.Duration) {
...
+	} else {
```
bwplotka, line 56: instead of else we can do `return` See https://thanos.io/contributing/coding-style-guide.md/#control-structure-prefer-early-returns-and-avoid-else

interceptors/logging/interceptors.go
```diff
@@ -47,40 +50,102 @@ func (c *reporter) PostCall(err error, duration time.Duration) {
...
+		if c.isServer {
+			logger.With(c.opts.durationFieldFunc(duration)...).Log(c.opts.levelFunc(code), fmt.Sprintf("request finished - response handled %s %s", c.kind, c.typ))
+		} else {
```
bwplotka, line 59: ditto

interceptors/logging/interceptors.go
```diff
@@ -47,40 +50,102 @@ func (c *reporter) PostCall(err error, duration time.Duration) {
 	if err != nil {
 		logger = logger.With("error", fmt.Sprintf("%v", err))
 	}
-	logger.With(c.opts.durationFieldFunc(duration)...).Log(c.opts.levelFunc(code), fmt.Sprintf("finished %s %s call", c.kind, c.typ))
+
+	if !c.opts.shouldLogRequest(interceptors.FullMethod(c.service, c.method), err) {
```
bwplotka, line 54: I don't get this, why we need this `if`? I think `shoullLog` controls this :thinking:

interceptors/logging/interceptors.go
```diff
@@ -47,40 +50,102 @@ func (c *reporter) PostCall(err error, duration time.Duration) {
...
+		logger.With(c.opts.durationFieldFunc(duration)...).Log(c.opts.levelFunc(code), fmt.Sprintf("finished %s %s call", c.kind, c.typ))
+	} else {
+		if c.isServer {
+			logger.With(c.opts.durationFieldFunc(duration)...).Log(c.opts.levelFunc(code), fmt.Sprintf("request finished - response handled %s %s", c.kind, c.typ))
+		} else {
+			logger.With(c.opts.durationFieldFunc(duration)...).Log(c.opts.levelFunc(code), fmt.Sprintf("response finished %s %s", c.kind, c.typ))
```
bwplotka, line 60: do we really need `response` `request` specialization here? also what you mean response finished? How response can finish itself? (: read of response can be finished...

I would just do as before 
```
	logger.With(c.opts.durationFieldFunc(duration)...).Log(c.opts.levelFunc(code), fmt.Sprintf("finished %s %s call", c.kind, c.typ))
```

bwplotka: What are the rationales for this/ (:

yashrsharma44: * Firstly, I don't want to break the initial Logging of the middlewares, if we don't enable the request logging, we can make sure that the earlier logging works as it was working before this pr.

* For the recording the completion of request/response, I chose to curate the message for the above. However, I do agree that the messages can be modified 😅

interceptors/logging/interceptors.go
yashrsharma44: I have added another decider to check if request logging is enabled.
My previous decider was breaking the tests and the initial API, so I thought that by using the decider as an option instead of my previous implementation, would break no changes, hence one more condition

bwplotka: Hm, think about what is the user experience here.

"If I configure `to log` I will have `finished .,.. call` log line."
"If I configure `to log request` I will have `request finished` if server side and `response finished` on caller. "
"If I configure `to log request, but to not log` I will have no log line"

Do you those three perceptions of current implementation will be ok for user? Because I think it can confuse users a lot. I think we can improve it much more :muscle: Think about how would configure this and what do you want to achieve. To me It's not `shouldLogRequest`. Focus on what you change here. Because to me, nothing is really needed. Why not just leaving `finished .. call`? (: 

What request finished` means and how `response` can be finished? `response` is  a noun?

yashrsharma44: > Do you those three perceptions of current implementation will be ok for user?

@bwplotka The confusion that arises out of the three implementation sounds fair. I think due to this confusion, it would make sense to allow one decider, rather than two currently in this PR

> Because to me, nothing is really needed. Why not just leaving finished .. call? (:

So should we skip out this PR altogether, because if the changes don't make sense and doesn't achieve as we wanted, it would make sense to skip out this PR 🙂

yashrsharma44: Currently, I configured the grpc middleware in Thanos, and the logs currently look like this before this pr gets merged - 
```
level=info name=sidecar-0 ts=2020-07-20T08:48:49.680407947Z caller=logger.go:31 service=gRPC/server component=sidecar system=grpc kind=server grpc.service=thanos.Store grpc.method=Info grpc.method_type=unary grpc.start_time=2020-07-20T14:18:49+05:30 grpc.request.deadline=2020-07-20T14:18:54+05:30 peer.address=127.0.0.1:57518 grpc.request.request-id=01EDNPJK0G6ZC1YHWNCVDMYVSN grpc.code=OK grpc.time_ms=0.054 finishedserverunarycall=(MISSING)
level=info name=sidecar-2 ts=2020-07-20T08:48:49.680587108Z caller=logger.go:31 service=gRPC/server component=sidecar system=grpc kind=server grpc.service=thanos.Store grpc.method=Info grpc.method_type=unary grpc.start_time=2020-07-20T14:18:49+05:30 grpc.request.deadline=2020-07-20T14:18:54+05:30 peer.address=127.0.0.1:37034 grpc.request.request-id=01EDNPJK0GX54BY58ACJ2XSZXW grpc.code=OK grpc.time_ms=0.044 finishedserverunarycall=(MISSING)
level=info name=sidecar-1 ts=2020-07-20T08:48:49.68244212Z caller=logger.go:31 service=gRPC/server component=sidecar system=grpc kind=server grpc.service=thanos.Store grpc.method=Info grpc.method_type=unary grpc.start_time=2020-07-20T14:18:49+05:30 grpc.request.deadline=2020-07-20T14:18:54+05:30 peer.address=127.0.0.1:59918 grpc.request.request-id=01EDNPJK0G94RMSP7CPEEB8J2D grpc.code=OK grpc.time_ms=0.034 finishedserverunarycall=(MISSING)
```

If the metadata required here looks comprehensive, then we can altogether skip merging this PR.

yashrsharma44:
Tests failing due to - https://github.com/grpc-ecosystem/go-grpc-middleware/issues/319

interceptors/logging/interceptors.go
```diff
@@ -45,42 +48,94 @@ func (c *reporter) PostCall(err error, duration time.Duration) {
 	code := c.opts.codeFunc(err)
 	logger = logger.With("grpc.code", code.String())
 	if err != nil {
-		logger = logger.With("error", fmt.Sprintf("%v", err))
+		logger = logger.With("grpc.error", fmt.Sprintf("%v", err))
+	}
+
+	logger = logger.With(c.opts.durationFieldFunc(duration)...)
+	logger = logger.With("protocol", "grpc")
+	logger.Log(c.opts.levelFunc(code), fmt.Sprintf("finished_%s_%s_call", c.kind, c.typ))
```
bwplotka, line 56: I think this is not what users might expect.

When you see some decision like this, especially such "controversial" (human readable streaming being with `_`) (: instead of going for it and spending time to adjust every PR, let's discuss first.

It's also clear like we need to common and consistent design on what such message we want to make this decision. I believe we could extend our proposal with such section. WDYT? (:

interceptors/logging/options.go
```diff
@@ -92,3 +92,7 @@ func DurationToDurationField(duration time.Duration) Fields {
 func durationToMilliseconds(duration time.Duration) float32 {
 	return float32(duration.Nanoseconds()/1000) / 1000
 }
+
+func DefaultRequestDecider(_ string, _ error) bool {
```
bwplotka, line 96: Is this used?

interceptors/logging/interceptors.go
```diff
@@ -45,42 +48,94 @@ func (c *reporter) PostCall(err error, duration time.Duration) {
...
+}
+
+// PostMsgSend logs the details of the servingObject that is flowing out of the rpc.
+// resp object wrt server.
+// Log the details of the first request, skip if the object is response.
+func (c *reporter) PostMsgSend(resp interface{}, err error, duration time.Duration) {
+	if !c.opts.shouldLog(interceptors.FullMethod(c.service, c.method), err) {
```
bwplotka, line 63: I think we might want bit more control, have some idea. (:

interceptors/logging/interceptors.go
yashrsharma44: Remove `_` from the messages

yashrsharma44: Add the key as `kind = c.kind` and `type = c.type`, and at the end log the `finished call`

yashrsharma44: `msg = finished call`

interceptors/logging/interceptors.go
yashrsharma44: Elaborate please

yashrsharma44: What are the options you wany for the decider?

yashrsharma44:
### Changes made uptil now - 
- [ ] Human Readable Messages - All the log entries are changed to use space instead of `_`.
- [ ] Using key-value pairs for storing the `kind`, `type` and `component`.
- [ ] Changed to `start call` and `finished call`.
### Changes that need to be made - 
- [ ] Add a design/doc section for using the API(Mentioned in my original TODO)
- [ ] Single Decider with multiple options, and an interface to configure the options in an easy way.

interceptors/logging/options.go
yashrsharma44: Must have forgotten to remove from the last change

yashrsharma44:
I would like to extend the `Decider` as using a struct instead of `ENUM`, because - 
* We would have `2^2` possibilities, so we need to keep `4` variables for checking the option of logging.
* We can use a struct for storing two variables to check if we want to log at start and end. This way, with minimal fuss, all the conditions are checked.

Interface of new Decider - 
```
type LogDecision struct {
     LogStarted, LogFinished bool
}
func NewLogDecision(LogStarted, LogFinished bool) *LogDecision {
     return &LogDecision {
          LogStarted : LogStarted,
          LogFinished : LogFinished,
     }
}
type Decider func (methodName string, err error) *LogDecision {
          // Have condition check
          return NewLogDecision(logStart, LogFinish)
}
```
WDYT?

interceptors/logging/interceptors.go
yashrsharma44: Changed

interceptors/logging/interceptors.go
```diff
@@ -22,65 +22,132 @@ func extractFields(tags tags.Tags) Fields {
...
+	logger                     Logger
 
 	kind string
 }
 
+func (c *reporter) addMetaDetails(logger Logger) Logger {
+	if c.isServer {
+		logger = logger.With("component", "server")
+	}
+	if !c.isServer {
```
kakkoyun, line 41: Why not else?

interceptors/logging/interceptors.go
```diff
@@ -22,65 +22,132 @@ func extractFields(tags tags.Tags) Fields {
...
+		logger = logger.With("component", "client")
+	}
+	// Attach the kind, protocol and type as a structured log entry.
+	logger = logger.With("kind", c.kind)
+	logger = logger.With("type", string(c.typ))
+	logger = logger.With("protocol", "grpc")
```
kakkoyun, line 47:
```suggestion
	logger = logger.With("kind", c.kind, "type", string(c.typ), "protocol", "grpc")
```

interceptors/logging/interceptors.go
```diff
@@ -22,65 +22,132 @@ func extractFields(tags tags.Tags) Fields {
...
-
+	logger = logger.With(extractFields(tags.Extract(c.ctx))...)
 	code := c.opts.codeFunc(err)
 	logger = logger.With("grpc.code", code.String())
 	if err != nil {
-		logger = logger.With("error", fmt.Sprintf("%v", err))
+		logger = logger.With("grpc.error", fmt.Sprintf("%v", err))
+	}
+	logger = logger.With(c.opts.durationFieldFunc(duration)...)
+	logger.Log(c.opts.levelFunc(code), "finished call")
```
kakkoyun, line 69:
```suggestion
	logger.Log(c.opts.levelFunc(code), "finished call", c.opts.durationFieldFunc(duration)...)
```

interceptors/logging/interceptors.go
```diff
@@ -22,65 +22,132 @@ func extractFields(tags tags.Tags) Fields {
...
+
+	// If the first message is logged skip the rest of the logging.
+	// If the serving object is response, skip the logging.
+	if c.firstResponseMessageLogged || !c.isServer {
+		return
+	}
+	c.firstResponseMessageLogged = true
+	logger := c.addMetaDetails(c.logger)
+	logger = logger.With(extractFields(tags.Extract(c.ctx))...)
+	logger = logger.With("grpc.recv.duration", duration.String())
```
kakkoyun, line 88:
```suggestion
	logger = logger.With("grpc.recv.duration", duration.String(), extractFields(tags.Extract(c.ctx))...)
```

interceptors/logging/interceptors.go
```diff
@@ -22,65 +22,132 @@ func extractFields(tags tags.Tags) Fields {
...
+	code := c.opts.codeFunc(err)
+	if err != nil {
+		logger = logger.With("grpc.error", fmt.Sprintf("%v", err))
 	}
-	logger.With(c.opts.durationFieldFunc(duration)...).Log(c.opts.levelFunc(code), fmt.Sprintf("finished %s %s call", c.kind, c.typ))
+	logger.With(c.opts.durationFieldFunc(duration)...).Log(c.opts.levelFunc(code), "started call")
```
kakkoyun, line 93:
```suggestion
	logger.Log(c.opts.levelFunc(code), "started call", c.opts.durationFieldFunc(duration)...)
```

interceptors/logging/interceptors.go
```diff
@@ -22,65 +22,132 @@ func extractFields(tags tags.Tags) Fields {
...
+	c.firstRequestMessageLogged = true
+	logger := c.addMetaDetails(c.logger)
+	logger = logger.With(extractFields(tags.Extract(c.ctx))...)
+	logger = logger.With("grpc.recv.duration", duration.String())
+	code := c.opts.codeFunc(err)
+
+	if err != nil {
+		logger = logger.With("grpc.error", fmt.Sprintf("%v", err))
+	}
+	logger.With(c.opts.durationFieldFunc(duration)...).Log(c.opts.levelFunc(code), "started call")
```
kakkoyun, line 118: ditto

interceptors/logging/interceptors.go
```diff
@@ -22,65 +22,132 @@ func extractFields(tags tags.Tags) Fields {
...
 
-func (c *reporter) PostMsgReceive(interface{}, error, time.Duration) {}
+	// If the first message for request is logged, skip the rest of the logging.
+	// If the serving object is a response, skip the logging.
+	if c.firstRequestMessageLogged || !c.isServer {
+		return
+	}
+	c.firstRequestMessageLogged = true
+	logger := c.addMetaDetails(c.logger)
+	logger = logger.With(extractFields(tags.Extract(c.ctx))...)
```
kakkoyun, line 111: ditto

kakkoyun:
@yashrsharma44 Ok, the interface is apparently different than I had assumed so you can ignore my comments or even better we can update the interface since this is a v2 branch.

interceptors/logging/interceptors.go
yashrsharma44: Yup can write that

interceptors/logging/interceptors.go
```diff
@@ -22,65 +22,132 @@ func extractFields(tags tags.Tags) Fields {
...
+
+// PostCall logs the server/method name details and error if any.
 func (c *reporter) PostCall(err error, duration time.Duration) {
 	if !c.opts.shouldLog(interceptors.FullMethod(c.service, c.method), err) {
 		return
 	}
 	if err == io.EOF {
 		err = nil
 	}
+	logger := c.addMetaDetails(c.logger)
```
bwplotka, line 60: Some of the things we can share with other calls instead of adding metas all the time.

interceptors/logging/interceptors.go
kakkoyun: Or 
```suggestion
	logger = logger.With("kind", c.kind).With("type", string(c.typ))..With("protocol", "grpc")
```

interceptors/logging/interceptors.go
```diff
@@ -22,65 +22,132 @@ func extractFields(tags tags.Tags) Fields {
...
 	// Get optional, fresh tags.
-	logger := c.logger.With(extractFields(tags.Extract(c.ctx))...)
-
+	logger = logger.With(extractFields(tags.Extract(c.ctx))...)
 	code := c.opts.codeFunc(err)
```
bwplotka, line 63: Those are also metadata we could put int some method to reuse.

interceptors/logging/interceptors.go
```diff
@@ -22,65 +22,132 @@ func extractFields(tags tags.Tags) Fields {
...
+// resp object wrt server.
+// Log the details of the first request, skip if the object is response.
+func (c *reporter) PostMsgSend(resp interface{}, err error, duration time.Duration) {
+	if !c.opts.shouldLog(interceptors.FullMethod(c.service, c.method), err) {
+		return
+	}
+
+	// If the first message is logged skip the rest of the logging.
+	// If the serving object is response, skip the logging.
+	if c.firstResponseMessageLogged || !c.isServer {
```
bwplotka, line 82: What about one bool flag? `startedCallLogged bool` ? (: for server and client the first PostMsgSend and PostMsgRecv being called "wins"

interceptors/logging/interceptors.go
bwplotka, line 93: I think those `c.opts.durationFieldFunc(duration)...` and `logger = logger.With("grpc.recv.duration", duration.String())` have exactly the same  valure

interceptors/logging/interceptors.go
yashrsharma44, line 118: Remove the duration function

interceptors/logging/interceptors.go
```diff
@@ -22,65 +22,132 @@ func extractFields(tags tags.Tags) Fields {
...
+	}
 
-func (c *reporter) PostMsgReceive(interface{}, error, time.Duration) {}
+	// If the first message for request is logged, skip the rest of the logging.
+	// If the serving object is a response, skip the logging.
+	if c.firstRequestMessageLogged || !c.isServer {
+		return
+	}
+	c.firstRequestMessageLogged = true
+	logger := c.addMetaDetails(c.logger)
```
bwplotka, line 110: I think all those things can be abstracted in to one function (so  PostMsgReceive and Send can just inoke `logStarted`)

interceptors/logging/interceptors.go
```diff
@@ -22,65 +22,132 @@ func extractFields(tags tags.Tags) Fields {
...
+}
 
 type reportable struct {
 	opts   *options
 	logger Logger
 }
 
-func (r *reportable) ServerReporter(ctx context.Context, _ interface{}, typ interceptors.GRPCType, service string, method string) (interceptors.Reporter, context.Context) {
-	return r.reporter(ctx, typ, service, method, KindServerFieldValue)
+func (r *reportable) ServerReporter(ctx context.Context, req interface{}, typ interceptors.GRPCType, service string, method string) (interceptors.Reporter, context.Context) {
```
bwplotka, line 126: Why we need `req` name? (:

interceptors/logging/interceptors_test.go
```diff
@@ -501,3 +537,271 @@ func (s *loggingCustomDeciderSuite) TestPingList_HasCustomDecider() {
 
 	require.Len(s.T(), s.logger.Lines(), 0) // Decider should suppress.
 }
+
+type requestLoggingSuite struct {
+	*baseLoggingSuite
+}
+
+func TestRequestLoggingSuite(t *testing.T) {
+
```
bwplotka, line 546:
```suggestion

```

interceptors/logging/interceptors_test.go
```diff
@@ -501,3 +537,271 @@ func (s *loggingCustomDeciderSuite) TestPingList_HasCustomDecider() {
...
+			tags.StreamServerInterceptor(tags.WithFieldExtractor(tags.CodeGenRequestFieldExtractor)),
+			logging.StreamServerInterceptor(s.logger, opts)),
+		middleware.WithUnaryServerChain(
+			tags.UnaryServerInterceptor(tags.WithFieldExtractor(tags.CodeGenRequestFieldExtractor)),
+			logging.UnaryServerInterceptor(s.logger, opts)),
+	}
+	suite.Run(t, s)
+}
+
+func (s *requestLoggingSuite) TestPing_RequestLogging() {
```
bwplotka, line 580: Do we need to duplicae all of this code? Sounds like we would rather extend the normal reporter test, as we kind of reuse the same thing.

interceptors/logging/interceptors_test.go
```diff
@@ -463,28 +493,34 @@ func (s *loggingCustomDeciderSuite) TestPingError_HasCustomDecider() {
...
 
-	clientLogLine := lines[1]
+	clientLogLine := lines[2]
 	assert.Equal(s.T(), logging.DEBUG, clientLogLine.lvl)
-	assert.Equal(s.T(), "finished client unary call", clientLogLine.msg)
+	assert.Equal(s.T(), "finished call", clientLogLine.msg)
 	clientFields := assertStandardFields(s.T(), logging.KindClientFieldValue, clientLogLine.fields, "PingError", interceptors.Unary)
 	clientFields.AssertNextFieldNotEmpty(s.T(), "grpc.start_time").
 		AssertNextFieldNotEmpty(s.T(), "grpc.request.deadline").
 		AssertNextField(s.T(), "grpc.code", "NotFound").
```
bwplotka, line 519: Reuse gives us power to reduce cognitive load which is resulted by complexity. And usually more lines of code means more stuff to read = complexity.

Thanks to reuse we did, we could .... reuse tests as well! So 5x less.

Now with our idea to have single in our code `addMeta` or `logStarted(` or even `logCall(isStarted, isSever)`, so we can focus on decider not on CHECKING this all the time: 
image

So again, on top 5x we 3x less to test. We just decide complexity by 15x

bwplotka:
Looks like we want decider to NOT have error and do adaptive logging base on error, using errCode to Log Level, :muscle: 


Looks like we are going into simpler functions, which means I think I would vote for `enums`

interceptors/logging/interceptors.go
yashrsharma44: Done

interceptors/logging/interceptors.go
yashrsharma44: Corrected

interceptors/logging/interceptors.go
yashrsharma44: Changed

interceptors/logging/interceptors.go
yashrsharma44: Yup, can skip that

interceptors/logging/interceptors.go
yashrsharma44: Changed

interceptors/logging/interceptors_test.go
yashrsharma44: Yup, extending the tests are better

interceptors/logging/interceptors_test.go
yashrsharma44: Yeah, reusing the test cases :muscle:

interceptors/logging/interceptors.go
yashrsharma44: Not possible with the current interface for `Log`

interceptors/logging/interceptors.go
yashrsharma44: Not possible with the current implementation of `Log`

yashrsharma44:
Once we fixate on the Decider implementation, I can start with the examples :nerd_face:

interceptors/logging/interceptors.go
```diff
@@ -25,34 +25,65 @@ type reporter struct {
...
+// resp object wrt server.
+// Log the details of the first request, skip if the object is response.
+func (c *reporter) PostMsgSend(_ interface{}, err error, duration time.Duration) {
+	if c.startCallLogged {
+		return
+	}
+	switch c.opts.shouldLog(interceptors.FullMethod(c.service, c.method)) {
+	case LogAllCall:
+		c.startCallLogged = true
+		// Log the start call
```
bwplotka, line 68: missing trailing period

bwplotka: and below

interceptors/logging/logging.go
```diff
@@ -14,10 +14,20 @@ import (
 	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors"
 )
 
+// ENUM values for enabling one of the three decisions for logging.
+type Decision int
+
+const (
+	NoLogCall Decision = iota
```
bwplotka, line 21: Might need some comments, but LGTM (:

interceptors/logging/logging.go
```diff
@@ -14,10 +14,20 @@ import (
 	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors"
 )
 
+// ENUM values for enabling one of the three decisions for logging.
+type Decision int
```
bwplotka, line 18: Comment has to start with `Decision` here

interceptors/logging/interceptors.go
yashrsharma44: We agreed on the decider

interceptors/logging/interceptors.go
yashrsharma44: Sure

interceptors/logging/logging.go
yashrsharma44: Done

interceptors/logging/interceptors.go
```diff
@@ -25,34 +25,65 @@ type reporter struct {
...
-	logger := c.logger.With(extractFields(tags.Extract(c.ctx))...)
+}
 
-	code := c.opts.codeFunc(err)
-	logger = logger.With("grpc.code", code.String())
-	if err != nil {
-		logger = logger.With("error", fmt.Sprintf("%v", err))
+// PostMsgSend logs the details of the servingObject that is flowing out of the rpc.
+// resp object wrt server.
+// Log the details of the first request, skip if the object is response.
```
bwplotka, line 60: obsolete comment I think

interceptors/logging/interceptors.go
```diff
@@ -25,34 +25,65 @@ type reporter struct {
...
+// req object wrt server.
+// Log the details of the request, skip if the object is response.
+func (c *reporter) PostMsgReceive(_ interface{}, err error, duration time.Duration) {
+	if c.startCallLogged {
+		return
+	}
+	switch c.opts.shouldLog(interceptors.FullMethod(c.service, c.method)) {
+	case LogAllCall:
+		c.startCallLogged = true
+		// Log the start call
```
bwplotka, line 83: Is this not clear enough btw ? ;p do we need those comments?

interceptors/logging/logging.go
```diff
@@ -14,10 +14,23 @@ import (
...
+// Decision defines rules for enabling start and end of logging.
+type Decision int
+
+const (
+	// NoLogCall - Logging is disabled.
+	NoLogCall Decision = iota
+	// LogFinishCall - Only finish logs of request is enabled.
+	LogFinishCall
+	// LogAllCall - Logging of start and end of request is enabled.
+	LogAllCall
```
bwplotka, line 26: I would be explicit and call it `LogStartAndFinishCall`

interceptors/logging/interceptors.go
yashrsharma44: Should I remove it?

yashrsharma44:
#319 comes again to haunt :ghost:

bwplotka:
I think that's something new: TestPayloadSuite/TestPingStream_LogsAllRequestsAndResponses: payload_test.go:166:

bwplotka:
https://github.com/grpc-ecosystem/go-grpc-middleware/pull/311/checks?check_run_id=907094454

yashrsharma44:
Yeah, its surprising because other tests pass, they should raise the same error :thinking:

yashrsharma44:
@bwplotka Can you run the tests again, to check if the error persists? Because I cannot reproduce it locally

yashrsharma44:
There seems to be a data race condition - https://github.com/grpc-ecosystem/go-grpc-middleware/pull/311/checks?check_run_id=907094454#step:5:161

kakkoyun:
@yashrsharma44 Is this race, related to your changes? Any work on fixing it?

yashrsharma44:
I tried to reproduce this error, but I am not getting any @kakkoyun

yashrsharma44:
@googlebot I consent.

bwplotka:
```
WARNING: DATA RACE
Read at 0x00c0003851e0 by goroutine 12:
  github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/logging_test.(*baseLoggingSuite).SetupTest()
      /home/runner/work/go-grpc-middleware/go-grpc-middleware/interceptors/logging/interceptors_test.go:141 +0xd1
  github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/logging_test.(*loggingPayloadSuite).SetupTest()
      <autogenerated>:1 +0x59
  github.com/stretchr/testify/suite.Run.func2()
      /home/runner/go/pkg/mod/github.com/stretchr/testify@v1.4.0/suite/suite.go:119 +0x5e1
  testing.tRunner()
      /opt/hostedtoolcache/go/1.14.4/x64/src/testing/testing.go:991 +0x1eb

Previous write at 0x00c0003851e0 by goroutine 96:
  github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/logging_test.(*baseMockLogger).Add()
      /home/runner/work/go-grpc-middleware/go-grpc-middleware/interceptors/logging/interceptors_test.go:101 +0x15b
  github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/logging_test.(*mockLogger).Log()
      /home/runner/work/go-grpc-middleware/go-grpc-middleware/interceptors/logging/interceptors_test.go:126 +0x28b
  github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/logging.logProtoMessageAsJson()
      /home/runner/work/go-grpc-middleware/go-grpc-middleware/interceptors/logging/payload.go:138 +0x1e5
  github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/logging.(*serverPayloadReporter).PostMsgSend()
      /home/runner/work/go-grpc-middleware/go-grpc-middleware/interceptors/logging/payload.go:36 +0x2b0
  github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors.(*monitoredServerStream).SendM

Goroutine 12 (running) created at:
  testing.(*T).Run()
      /opt/hostedtoolcache/go/1.14.4/x64/src/testing/testing.go:1042 +0x660
  github.com/stretchr/testify/suite.runTests()
      /home/runner/go/pkg/mod/github.com/stretchr/testify@v1.4.0/suite/suite.go:151 +0xe3
  github.com/stretchr/testify/suite.Run()
      /home/runner/go/pkg/mod/github.com/stretchr/testify@v1.4.0/suite/suite.go:138 +0x6af
  github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/logging_test.TestPayloadSuite()
      /home/runner/work/go-grpc-middleware/go-grpc-middleware/interceptors/logging/payload_test.go:57 +0xbdc
  testing.tRunner()
      /opt/hostedtoolcache/go/1.14.4/x64/src/testing/testing.go:991 +0x1eb
Goroutine 96 (finished) created at:
  google.golang.org/grpc.(*Server).serveStreams.func1()
      /home/runner/go/pkg/mod/google.golang.org/grpc@v1.19.0/server.go:688 +0xb8
  google.golang.org/grpc/internal/transport.(*http2Server).operateHeaders()
      /home/runner/go/pkg/mod/google.golang.org/grpc@v1.19.0/internal/transport/http2_server.go:419 +0x14e0
  google.golang.org/grpc/internal/transport.(*http2Server).HandleStreams()
      /home/runner/go/pkg/mod/google.golang.org/grpc@v1.19.0/internal/transport/http2_server.go:459 +0x31e
  google.golang.org/grpc.(*Server).serveStreams()
      /home/runner/go/pkg/mod/google.golang.org/grpc@v1.19.0/server.go:686 +0x1b5
  google.golang.org/grpc.(*Server).handleRawConn.func1()
      /home/runner/go/pkg/mod/google.golang.org/grpc@v1.19.0/server.go:648 +0x50
==================
```

yashrsharma44:
I know it's been a long time, but can I have another round of reviews :grin: ? cc @bwplotka

interceptors/logging/interceptors.go
```diff
@@ -25,34 +25,55 @@ type reporter struct {
...
 
-	opts   *options
-	logger Logger
-
-	kind string
+func (c *reporter) logMessage(logger Logger, err error, msg string, duration time.Duration) {
+	code := c.opts.codeFunc(err)
+	logger = logger.With("grpc.code", code.String())
+	if err != nil {
+		logger = logger.With("grpc.error", fmt.Sprintf("%v", err))
```
bwplotka, line 38:
```suggestion
		logger = logger.With("grpc.error", err.Error())
```

bwplotka (approved):
Tiny nit only, we can fix that later (: 

LGTM, Thanks! :+1:

interceptors/logging/logging.go
```diff
@@ -47,12 +60,12 @@ func DefaultErrorToCode(err error) codes.Code {
 }
 
 // Decider function defines rules for suppressing any interceptor logs
-type Decider func(fullMethodName string, err error) bool
+type Decider func(fullMethodName string) Decision
```
yashrsharma44, line 63: Would have been nice not to have skipped the error parameter here. The logging can also be configured using the error as well,(since status code can be derived from the response error).

Issue #277, opened 2020-03-02
Add opentemeletry-go tracing support on top of  `opentracing`.


bwplotka:
https://github.com/open-telemetry/opentelemetry-go/blob/master/api/trace/api.go

bwplotka:
Actually the current idea is to support both interfaces (: in `v2` branch. See: https://github.com/grpc-ecosystem/go-grpc-middleware/issues/275

bwplotka:
Acceptance Criteria:

* Users can use tracing interceptor with opentelemetry-go APIs too. 

NOTE:  
* Opentelemetry project is not Stable yet, APIs are changing that's why another provider and not replacing opentracing

OpenTelemetry is bit confusing, there are many abstractions. In practice I found it useful to create simpler wrapper for thousands of modules they have: * I would suggest using some abstraction like https://github.com/AnaisUrlichs/observe-argo-rollout/tree/main/app/tracing 

I don't think we need to use this code, but we might want to add that for testing, or your own experimentation purposes (who ever is on this issue)

bwplotka:
This is easier than we thought thanks of OpenTelemetry.

https://github.com/grpc-ecosystem/go-grpc-middleware/pull/543 changes v2 to *remove* tracing completely, since two interceptors are allowed:

* Tracing:
  * (external) [`otelgrpc`](https://go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc) - official OpenTelemetry tracing interceptors as used in [example](examples).
  * (external) [`opentracing`](https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware@v1.4.0/tracing/opentracing) - deprecated [OpenTracing](http://opentracing.io/) client-side and server-side interceptors if you still need it!

PR #312, opened 2020-06-30
Refactor tracing implements

XSAM:
Split from #275

- Add tracing interceptor.
- Add opentracing provider.
- Add opentelemetry provider.


XSAM:
@bwplotka I need some review to confirm the tracing interface since it implements parts of `interceptors.Reporter`. 😁

interceptors/tracing2/interceptors.go
```diff
@@ -0,0 +1,70 @@
+package tracing2
```
bwplotka, line 1: Why not modifying `tracing` directly?

XSAM: It makes the code easier to review.

I will rename it back to `tracing` in the last commit.

XSAM:
Well, `opentracing` and `OTEL (opentelemtry)` do have a familiar interface of span to record events. 

`opentracing`: LogFields, `OTEL`: AddEvent.

But the problem is the behavior from two libs are not the same while handling `PostMsgSend` and `PostMsgReceive`.

For instance, when invoking `PostMsgSend`, `opentracing` will do nothing with `LogFields` since it doesn't require to record gRPC message info, but `OTEL` will record that info via `AddEvent`.

Furthermore, attributes which been using in the event are different.

`OTEL`:

```go
span.AddEvent(ctx, "message",
	kv.Key("message.type").String("SENT"),
	kv.Key("message.id").Int(id),
	kv.Key("message.uncompressed_size").Int(proto.Size(body)),
)
```

So, I think there are two solutions to this.

1. Add `AddEvent` function for span interface. But it will force `opentracing` to add the same key and attributes to events like `OTEL`, or `opentracing` have to do some weird mapping to drop these non-standard events.

2. Add `PostMsgSend` and `PostMsgReceive` for span interface. Weird but more flexible.

-----

- https://github.com/opentracing/specification/blob/master/semantic_conventions.md
- https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/rpc.md#events

bwplotka:
Hi @XSAM Let's get back to this!

I feel like option number 2 is too similar too reporter, so kind of no point in having separate span interface, no? (: 

That's why we have two options I think

1. as your 1, we could try that.

3.  Actually have no common interface among those, just have 2 different providers/opentelemetry` and `providers/opentracing` ^^

XSAM:
You're right, option 2 is too similar with `interceptors.Reporter`.

Now, I use option 1 as a solution.

yashrsharma44:
Seems like this PR would address #277 . @XSAM can you let us know if you are interested in working on rebasing this PR and meeting the acceptance criteria? Would be happy to take over this PR, otherwise 😄 

Related Issue - https://github.com/grpc-ecosystem/go-grpc-middleware/issues/277

XSAM:
@yashrsharma44 I will continue to work on this PR. Thanks for the reminder. 😄

XSAM:
I updated this PR. @yashrsharma44 @bwplotka PTAL ;)

XSAM:
Any updates on this one?

yashrsharma44:
Hi @XSAM!
Extremely sorry for the late review, I was busy with my internship requirements, so wasn't able to take a look at the PR.

* Can you split the changes for `opentracing` and `opentelemetry` in a separate PR?
* Also can you rebase your PR with the latest changes?

bwplotka:
I think it's actually reviewable, in one place as those are new things mostly. But it will take time. 

Still, I think we should do it before v2?

bwplotka:
Are you around @XSAM to work on it (rebase etc)? 

Also I am removing tracing here, to avoid modifications there that will have to be there as part of this Pr anyway (https://github.com/grpc-ecosystem/go-grpc-middleware/pull/394#issuecomment-894643059)

XSAM:
> Are you around @XSAM to work on it (rebase etc)?
> 
> Also I am removing tracing here, to avoid modifications there that will have to be there as part of this Pr anyway ([#394 (comment)](https://github.com/grpc-ecosystem/go-grpc-middleware/pull/394#issuecomment-894643059))

Sorry for the late response, I was busy recently.

I'm still working on it; will try to split these changes for clarity.

PR #313, opened 2020-07-08
Do not stop retrying based on earlier good message from the stream

kartlee:
- Usually, the pattern is to establish a stream and read messages in the loop until io.EOF. If the server becomes unavailable after sending a good message in the stream, we stop retrying in the middleware based on an earlier good message.  This forces the caller to check for codes.Unavailable and initiate establishing the stream and read again. The patch handles this case inside the middleware.

### DIFF
--- a/retry/retry.go
+++ b/retry/retry.go
@@ -136,7 +136,6 @@ func StreamClientInterceptor(optFuncs ...CallOption) grpc.StreamClientIntercepto
 type serverStreamingRetryingStream struct {
 	grpc.ClientStream
 	bufferedSends []interface{} // single message that the client can sen
-	receivedGood  bool          // indicates whether any prior receives were successful
 	wasClosedSend bool          // indicates that CloseSend was closed
 	parentCtx     context.Context
 	callOpts      *options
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
I'm not convinced this is as straightforward a change as it looks. Why was this logic added if we can just remove it? Could you please add some tests that exhibits the issue and why this change is necessary?

kartlee:
Sure,  I will see the existing test framework and try to add one. I looked at the history of the line of code and didn't get an explanation of why wasGood was added in the first place. 

0) client gets a server stream
1) loop reading the stream to get messages until an error is received.
2) The server sends a good message and restarts.
3)  The client gets a good message and comes out of the loop without retrying in backoff to wait for the server to come back.

-Karthik

bwplotka:
But it's a good point to take a look here what's going on. The problem is that if it was required, there was no tests for it (as CI is green), so it might be good idea to revisit this.

Also do you mind changing `v2` branch? @kartlee

kartlee:
> But it's a good point to take a look here what's going on. The problem is that if it was required, there was no tests for it (as CI is green), so it might be good idea to revisit this.
> 
> Also do you mind changing `v2` branch? @kartlee

For v2 -> https://github.com/grpc-ecosystem/go-grpc-middleware/pull/323

Thanks for approving the change to master.

bwplotka:
Let's make exception here, normally we should slowly stop updating v1

bwplotka:
Anyway, thanks for both changes!

PR #316, opened 2020-07-16
inline localhost certificate into go file

### LINKED ISSUE #285: embed localhost.crt in test go files
It's not currently possible to vendor the testing suite with `go mod`. This is because `go mod` does not import repos, it only imports go packages. The testing suite makes references to `testing/certs/localhost.{crt,key}` which are not imported.

bmon:
This change allows library clients correctly run test suites when vendoring modules. Before this change, the certificate files would be pruned when vendoring, causing the tests to fail or deadlock.

Also included in this PR is a timeout for `InterceptorTestSuite.SetupSuite` which prevents a deadlock that occurs if an error is encountered when establishing the server.

resolves #285


johanbrandhorst:
Thanks for your contribution! Could you merge this to v2 as well?

PR #318, opened 2020-07-16
inline localhost certificate into go file

bmon:

## Changes

This change allows library clients correctly run test suites when vendoring modules. Before this change, the certificate files would be pruned when vendoring, causing the tests to fail or deadlock.

Also included in this PR is a timeout for InterceptorTestSuite.SetupSuite which prevents a deadlock that occurs if an error is encountered when establishing the server.

## Verification


go.sum
```diff
@@ -13,6 +13,7 @@ github.com/golang/mock v1.1.1/go.mod h1:oTYuIxOrZwtPieC+H1uAHpcLFnEyAGVDL/k47Jfb
 github.com/golang/protobuf v1.2.0/go.mod h1:6lQm79b+lXiMfvg/cZm0SGofjICqVBUtrP5yJMmIC1U=
 github.com/golang/protobuf v1.3.2 h1:6nsPYzhq5kReh6QImI3k5qWzO4PEbvbIW2cwSfR/6xs=
 github.com/golang/protobuf v1.3.2/go.mod h1:6lQm79b+lXiMfvg/cZm0SGofjICqVBUtrP5yJMmIC1U=
+github.com/grpc-ecosystem/go-grpc-middleware v1.2.0 h1:0IKlLyQ3Hs9nDaiK5cSHAGmcQEIC8l2Ts1u6x5Dfrqg=
```
johanbrandhorst, line 16: I think this was probably added in error, could you try running `go mod tidy`?

bmon: Right, sorry about that!

johanbrandhorst:
Looks like we've got a flaky test, I'll rerun it.

PR #320, opened 2020-07-24
Update streaming interceptor example

G07cha:
Return new stream with wrapped context instead of original so changes will be passed further the chain


G07cha:
@googlebot I signed it!

johanbrandhorst:
Thanks for your contribution! Could you please see if this applies to the v2 documentation as well?

G07cha:
@johanbrandhorst good point! I've cherry-picked commit in separate PR(#322) as well.

PR #322, opened 2020-07-24
Update streaming interceptor example

G07cha:
Return new stream with wrapped context instead of original so changes will be passed further the chain

This change is already landed in master branch in #320


johanbrandhorst:
Thanks for your contribution!

PR #325, opened 2020-08-14
generate TLS test certs with SAN for go1.15

### LINKED ISSUE #324: unit tests fail with go1.15 due to X.509 CommonName deprecation
`go test ./...` in the repo root works OK with go1.14.7 but fails with go1.15.  It also works fine when run as `GODEBUG=x509ignoreCN=0 go test ./...` which indicates that the failure is related to the X.509 CommonName deprecation in 1.15 [1].

The CommonName-related errors are better visible if you comment out the blocking option in 
https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/testing/interceptor_suite.go#L105 - changing that line to:
`newDialOpts := dialOpts` and running `go test ./...` produces multiple errors like
```
--- FAIL: TestAuthTestSuite (0.02s)
    --- FAIL: TestAuthTestSuite/TestStream_BadAuth (0.00s)
        auth_test.go:127:
                Error Trace:    auth_test.go:127
                Error:          Received unexpected error:
                                rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: x509: certificate relies on legacy Common Name field, use SANs or temporarily enable Common Name matching with GODEBUG=x509ignoreCN=0"
                Test:           TestAuthTestSuite/TestStream_BadAuth
                Messages:       should not fail on establishing the stream
````

Full dump:
[grpc-middleware.txt](https://github.com/grpc-ecosystem/go-grpc-middleware/files/5076961/grpc-middleware.txt)

We also have failures in the packages that depend on github.com/grpc-ecosystem/go-grpc-middleware with go1.15 that go away when building with go1.14.7 or using `GODEBUG=x509ignoreCN=0 go test -v ./...`.

[1] https://golang.org/doc/go1.15#commonname

### PR DESCRIPTION
**Pull request by dmitris**
Fixes #324. The certificates need to have SAN fields
instead of relying on the CommonName due to
X.509 CommonName deprecation in go1.15.

Reference: https://golang.org/doc/go1.15#commonname

/cc @johanbrandhorst

### CODE REVIEW

**Comment by dmitris** [comment]
@googlebot I consent.

**Comment by dmitris** [comment]
@googlebot I consent

**Comment by dmitris** [comment]
@googlebot I consent

File: testing/certs/gen_cert.sh [hunk 1]
```diff
@@ -1,7 +1,10 @@
 #!/bin/bash
 # Regenerate the self-signed certificate for local host.
 
-openssl req -x509 -sha256 -nodes -newkey rsa:2048 -days 3650 -keyout localhost.key -out localhost.crt
+openssl req -x509 -sha256 -nodes -newkey rsa:2048 -days 3650 \
+	-keyout localhost.key -out localhost.crt \
+	-subj "/C=US/ST=California/L=Sunnivale/O=/CN=www.example.com" \
```
maditya, line 6:
```suggestion
	-subj "/C=US/ST=California/L=Sunnyvale/O=/CN=www.example.com" \
```

dmitris:
@googlebot I fixed it

johanbrandhorst:
Hi @dmitris, thanks for this initiative. I think now would be a good time to replace this certificate generation code altogether, and use the Go x509 stack to generate certs on the fly, so we never have to worry about expiry. Would you mind implementing something like this for the tests that need it? You can look at https://golang.org/src/crypto/tls/generate_cert.go for inspiration.

johanbrandhorst:
Hi @dmitris, did you see my last comment?

testing/interceptor_suite.go
```diff
@@ -143,3 +161,76 @@ func (s *InterceptorTestSuite) TearDownSuite() {
...
+	f, err := os.OpenFile("localhost.go", os.O_RDWR|os.O_TRUNC|os.O_CREATE, 0644)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+	if err := t.Execute(f, data); err != nil {
+		return err
+	}
+	return nil
+}
```
johanbrandhorst, line 195: I think we could replace this whole function with one that generates a certificate and key and then returns the in-memory representation. Now that we're doing it in the application we shouldn't need to use the filesystem at all. See something like https://github.com/johanbrandhorst/certify/blob/master/issuers/vault/vault_suite_test.go#L255.

johanbrandhorst: We can then use https://pkg.go.dev/google.golang.org/grpc@v1.31.1/credentials?tab=doc#NewClientTLSFromCert and https://pkg.go.dev/google.golang.org/grpc@v1.31.1/credentials?tab=doc#NewServerTLSFromCert.

dmitris: removed the filesystem usage - had to use this snippet instead of `NewClientTLSFromCert` - similar to https://github.com/grpc/grpc-go/blob/v1.31.1/credentials/tls.go#L191
```
		cp := x509.NewCertPool()
		if !cp.AppendCertsFromPEM(testcert.CertPEM()) {
			s.T().Fatal("failed to append certificate")
		}
		creds := credentials.NewTLS(&tls.Config{ServerName: "localhost", RootCAs: cp})
```

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution! Could you please cherry-pick this against the v2 branch?

PR #329, opened 2020-08-31
Client unary interceptor timeout

instabledesign:
👋  Hi, 

what do you think about this addition?

Do you think it must purpose a most open way to determine the timeout (base on request or other input) in order to have a better control.

Let me know if this functionality already exist, and where?

Thank's for reading

### DIFF
--- a/timeout/doc.go
+++ b/timeout/doc.go
@@ -1,0 +1,11 @@
+// Copyright 2017 David Ackroyd. All Rights Reserved.
+// See LICENSE for licensing terms.
+
+/*
+`grpc_timeout` are interceptors that timeout for gRPC client calls.
+
+Client Side Timeout Middleware
+
+Please see examples for simple examples of use.
+*/
+package grpc_timeout
--- a/timeout/examples_test.go
+++ b/timeout/examples_test.go
@@ -1,0 +1,42 @@
+// Copyright 2017 David Ackroyd. All Rights Reserved.
+// See LICENSE for licensing terms.
+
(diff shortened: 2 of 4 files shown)


johanbrandhorst:
This is a cool idea, I don't think there is anything like it. We're currently working on finishing off the v2 branch though, and I'm wondering if you'd like to submit this against v2 instead? Also, one of the file contains a copyright header, and I'm not sure that will be possible to contribute to this repository. Are you the owner of the copyright?

instabledesign:
I thank,
to be honest i just copy paste the copyright from the [recovery](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/recovery/interceptors.go) in order to have a PR as much similar as i can.
I am not the owner of this copyright comment bloc but the code is mine. 
Should i just put my name, or delete the complete copyright?

johanbrandhorst:
You should just remove it, it is not necessary.

timeout/doc.go
```diff
@@ -0,0 +1,11 @@
+// Copyright 2017 David Ackroyd. All Rights Reserved.
+// See LICENSE for licensing terms.
+
```
johanbrandhorst, line 3: Remove this, if it is not subject to this license.

johanbrandhorst:
I'd like users to move to v2, so maybe we shouldn't merge this to v1? What do you think?

instabledesign:
Yeah let's encourage them to upgrade to v2, feel free to close this one 😄

PR #330, opened 2020-09-01
v2: Client unary interceptor timeout on v2 branch

instabledesign:
👋  Hi, 
this PR is related to the first one on the v1 #329 
It was reworked for the v2 branch

### DIFF
--- a/interceptors/timeout/examples_test.go
+++ b/interceptors/timeout/examples_test.go
@@ -11,7 +11,7 @@ import (
 )
 
 // Initialization shows an initialization sequence with a custom client request timeout.
-func Example_initialization() error {
+func Example_initialization() {
 	clientConn, err := grpc.Dial(
 		"ServerAddr",
 		grpc.WithUnaryInterceptor(
@@ -20,14 +20,14 @@ func Example_initialization() error {
 		),
 	)
 	if err != nil {
-		return err
+		log.Fatal(err)
 	}
 
 	// Initialize your grpc service with connection
 	testServiceClient := testpb.NewTestServiceClient(clientConn)
 	resp, err := testServiceClient.PingEmpty(context.TODO(), &testpb.Empty{})
 	if err != nil {
-		return err
+		log.Fatal(err)
 	}
 
 	// Use grpc response value
--- a/interceptors/timeout/examples_test.go
+++ b/interceptors/timeout/examples_test.go
@@ -11,7 +11,7 @@ import (
 )
 
 // Initialization shows an initialization sequence with a custom client request timeout.
-func Example_initialization() {
+func Example_initialization() error {
 	clientConn, err := grpc.Dial(
 		"ServerAddr",
 		grpc.WithUnaryInterceptor(
@@ -19,19 +19,15 @@ func Example_initialization() {
 			timeout.TimeoutUnaryClientInterceptor(20*time.Millisecond),
 		),
 	)
-
-	// Handle connection error
 	if err != nil {
-		panic(err)
+		return err
 	}
 
 	// Initialize your grpc service with connection
 	testServiceClient := testpb.NewTestServiceClient(clientConn)
 	resp, err := testServiceClient.PingEmpty(context.TODO(), &testpb.Empty{})
-
-	// Handle request error
 	if err != nil {
-		panic(err)
+		return err
 	}
 
 	// Use grpc response value
--- a/interceptors/timeout/examples_test.go
+++ b/interceptors/timeout/examples_test.go
@@ -2,7 +2,7 @@ package timeout_test
 
 import (
 	"context"
-	"fmt"
+	"log"
 	"time"
 
 	"github.com/grpc-ecosystem/go-grpc-middleware/v2/grpctesting/testpb"
@@ -35,5 +35,5 @@ func Example_initialization() {
 	}
 
 	// Use grpc response value
-	fmt.Println(resp.Value)
+	log.Println(resp.Value)
 }
--- a/interceptors/timeout/examples_test.go
+++ b/interceptors/timeout/examples_test.go
@@ -1,0 +1,39 @@
+package timeout_test
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/grpctesting/testpb"
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/timeout"
+	"google.golang.org/grpc"
+)
+
+// Initialization shows an initialization sequence with a custom client request timeout.
+func Example_initialization() {
+	clientConn, err := grpc.Dial(
+		"ServerAddr",
+		grpc.WithUnaryInterceptor(
+			// Set your client request timeout
+			timeout.TimeoutUnaryClientInterceptor(20*time.Millisecond),
+		),
+	)
+
+	// Handle connection error
+	if err != nil {
+		panic(err)
+	}
+
+	// Initialize your grpc service with connection
+	testServiceClient := testpb.NewTestServiceClient(clientConn)
+	resp, err := testServiceClient.PingEmpty(context.TODO(), &testpb.Empty{})
+
+	// Handle request error
+	if err != nil {
+		panic(err)
+	}
+
+	// Use grpc response value
+	fmt.Println(resp.Value)
+}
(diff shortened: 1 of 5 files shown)


eko (approved):

interceptors/timeout/examples_test.go
```diff
@@ -0,0 +1,39 @@
...
+	resp, err := testServiceClient.PingEmpty(context.TODO(), &testpb.Empty{})
+
+	// Handle request error
+	if err != nil {
+		panic(err)
+	}
+
+	// Use grpc response value
+	log.Println(resp.Value)
+}
```
johanbrandhorst, line 39: Lets apply some best practices to this code.
```suggestion
func Example_initialization() error {
	clientConn, err := grpc.Dial(
		"ServerAddr",
		grpc.WithUnaryInterceptor(
			// Set your client request timeout
			timeout.TimeoutUnaryClientInterceptor(20*time.Millisecond),
		),
	)
	if err != nil {
		return err
	}

	// Initialize your grpc service with connection
	testServiceClient := testpb.NewTestServiceClient(clientConn)
	resp, err := testServiceClient.PingEmpty(context.TODO(), &testpb.Empty{})
	if err != nil {
		return err
	}

	// Use grpc response value
	log.Println(resp.Value)
}
```

johanbrandhorst (commented):
Thanks for this PR! Could you just update the example?

gusmin (approved):

instabledesign:
what do you think if we delegate the choice/computation of the timeout through a function in order to have method specific timeout
```go
func TimeoutUnaryClientInterceptor(timeoutProvider func(string) time.Duration) grpc.UnaryClientInterceptor {
	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
		timedCtx, cancel := context.WithTimeout(ctx, timeoutProvider(method))
		defer cancel()
		return invoker(timedCtx, method, req, reply, cc, opts...)
	}
}
```

johanbrandhorst:
> what do you think if we delegate the choice/computation of the timeout through a function in order to have method specific timeout
> 
> ```go
> func TimeoutUnaryClientInterceptor(timeoutProvider func(string) time.Duration) grpc.UnaryClientInterceptor {
> 	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
> 		timedCtx, cancel := context.WithTimeout(ctx, timeoutProvider(method))
> 		defer cancel()
> 		return invoker(timedCtx, method, req, reply, cc, opts...)
> 	}
> }
> ```

If the user wants method specific timeouts they can just do that when calling an RPC. I think it makes sense to have a per-grpc-client timeout configuration, so lets keep it as is.

johanbrandhorst (changes requested):
Another thought; could you implement this in terms of the `ClientReportable` interface, like the other interceptors do? CC @bwplotka.

instabledesign:
I don't find this way clearer than previous 
and this one is "not really" a "Reporter"

Something similar than 
```go
type TimeoutUnaryClientReportable struct {
	timeout time.Duration
}

func (t *TimeoutUnaryClientReportable) ClientReporter(ctx context.Context, reqProtoOrNil interface{}, typ interceptors.GRPCType, service string, method string) (interceptors.Reporter, context.Context) {
	timedCtx, cancel := context.WithTimeout(ctx, t.timeout)
	return &TimeoutReporter{cancel: cancel}, timedCtx
}

type TimeoutReporter struct {
	cancel context.CancelFunc
}

func (t *TimeoutReporter) PostMsgReceive(replyProto interface{}, err error, recvDuration time.Duration) {
	t.cancel()
}

func (t *TimeoutReporter) PostCall(err error, rpcDuration time.Duration) {

}

func (t *TimeoutReporter) PostMsgSend(reqProto interface{}, err error, sendDuration time.Duration) {

}
```

johanbrandhorst:
I agree this is less clear. What do you think @bwplotka?

interceptors/timeout/timeout.go
```diff
@@ -0,0 +1,17 @@
...
+	"time"
+
+	"google.golang.org/grpc"
+)
+
+// TimeoutUnaryClientInterceptor returns a new unary client interceptor that sets a timeout on the request context.
+func TimeoutUnaryClientInterceptor(timeout time.Duration) grpc.UnaryClientInterceptor {
+	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+		timedCtx, cancel := context.WithTimeout(ctx, timeout)
+		defer cancel()
```
bwplotka, line 14: Just curious, how this is different to setting timeout on gRPC client side, or passing context with timeout?

bwplotka (commented):
Not sure if this is really needed, let's double check. If it's then I am fine non  reporter version of it - it's quite simple.

interceptors/timeout/timeout.go
instabledesign: this middleware just ensure that all grpc call has a "default timeout".
in other hand i have commit c747443231c71873c05b7b9ec2d512e14cf45b9c a fix where the incoming context already has timeout (it will keep it)

bwplotka: Cool: https://github.com/grpc/grpc-go/issues/2130

bwplotka (approved):
LGTM! 

Some tests are not working though

interceptors/timeout/examples_test.go
```diff
@@ -0,0 +1,35 @@
...
+		grpc.WithUnaryInterceptor(
+			// Set your client request timeout
+			timeout.TimeoutUnaryClientInterceptor(20*time.Millisecond),
+		),
+	)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Initialize your grpc service with connection
```
bwplotka, line 26: missing trailing period on all comments here (:

instabledesign:
I am not really fan when test depend on timing
I just increase the timing to work on OSX too 
it seems the github osx was slowest that my personal computer

interceptors/timeout/timeout.go
```diff
@@ -0,0 +1,20 @@
...
+
+	"google.golang.org/grpc"
+)
+
+// TimeoutUnaryClientInterceptor returns a new unary client interceptor that sets a timeout on the request context.
+func TimeoutUnaryClientInterceptor(timeout time.Duration) grpc.UnaryClientInterceptor {
+	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+		if _, ok := ctx.Deadline(); ok {
+			return invoker(ctx, method, req, reply, cc, opts...)
+		}
```
johanbrandhorst, line 15: I don't think this is the right design, adding a new context timeout to a context with an existing timeout isn't going to overwrite it, they will just both be active. I think we can remove this.

bwplotka:
Yea, ideally we want to remove all time-based tests, it's kind of legacy.
Help wanted (:

instabledesign:
But the main purpose of this middleware is to deal with time.
Have you any suggestion about how it can be done?
I think used channel, but it doesn't fix the middleware timer

johanbrandhorst:
Not sure there's much choice here since we're using `context.WithTimeout` directly, and we can't inject a custom clock or anything. If we wanted to make it testable you'd make the context creator configurable, so that in a test you could replace it with something deterministic, while in production use it'd just use `context.WithTimeout`. Personally I'm not sure that level of indirection is necessary. Can we just use single millisecond sleeps and timeouts?

instabledesign:
i agree with you, i am not really fan of doing such things only for testability purpose.

But if it stabilize tests suite, i can do a pattern option to customize the `context.WithTimeout` creation.

johanbrandhorst:
Nah that is only going to risk users abusing it. Unless @bwplotka has a concrete suggestion I think we need to do a timing test here.

bwplotka (approved):
Make sense! 

Can you add entry to README.md as well? 🤗 


Thanks 💪🏽 Sorry for lag in review! Let's get v2 released!

bwplotka:
Tests has to pass as well (:

instabledesign:
Hi!!
I rebase from v2, and update tests accordling to v2 changes
Hope its enough

instabledesign:
@googlebot I consent.

bwplotka (approved):
Ultra simple, but great! (: Thanks. LGTM

PR #336, opened 2020-09-09
remove 1.11 and 1.12 from travis config

dmitris:
PR removes 1.11 and 1.12 to avoid the build error due to the grpc dependency.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 sudo: false
 language: go
 go:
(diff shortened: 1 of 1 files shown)


johanbrandhorst (approved):
These releases are no longer supported by the Go team.

johanbrandhorst:
Thanks for your contribution! Could you cherry pick this against v2 please?

Issue #319, opened 2020-07-16
v2: Race condition in logging interceptor test


johanbrandhorst:
While merging an unrelated fix we had a test failure:

https://github.com/grpc-ecosystem/go-grpc-middleware/pull/318/checks?check_run_id=877450083

The relevant part is this:

```
2020-07-16T11:56:27.0723162Z === RUN   TestPayloadSuite/TestPing_LogsBothRequestAndResponse
2020-07-16T11:56:27.0723277Z ==================
2020-07-16T11:56:27.0723533Z WARNING: DATA RACE
2020-07-16T11:56:27.0723640Z Read at 0x00c000402bc0 by goroutine 12:
2020-07-16T11:56:27.0724126Z   github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/logging_test.(*baseLoggingSuite).SetupTest()
2020-07-16T11:56:27.0724600Z       /home/runner/work/go-grpc-middleware/go-grpc-middleware/interceptors/logging/interceptors_test.go:141 +0xd1
2020-07-16T11:56:27.0724934Z   github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/logging_test.(*loggingPayloadSuite).SetupTest()
2020-07-16T11:56:27.0725124Z       <autogenerated>:1 +0x59
2020-07-16T11:56:27.0725237Z   github.com/stretchr/testify/suite.Run.func2()
2020-07-16T11:56:27.0725471Z       /home/runner/go/pkg/mod/github.com/stretchr/testify@v1.4.0/suite/suite.go:119 +0x5e1
2020-07-16T11:56:27.0725607Z   testing.tRunner()
2020-07-16T11:56:27.0725764Z       /opt/hostedtoolcache/go/1.14.4/x64/src/testing/testing.go:991 +0x1eb
2020-07-16T11:56:27.0725834Z 
2020-07-16T11:56:27.0726118Z Previous write at 0x00c000402bc0 by goroutine 96:
2020-07-16T11:56:27.0726452Z   github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/logging_test.(*baseMockLogger).Add()
2020-07-16T11:56:27.0726853Z       /home/runner/work/go-grpc-middleware/go-grpc-middleware/interceptors/logging/interceptors_test.go:101 +0x15b
[...]

PR #354, opened 2020-10-06
Bug fix for data race

yashrsharma44:
Changes made in this PR:
- [x] Fix the data race bug - I have used locks wherever `append` is called.
- [x] Updated the importing version of `grpc-middleware` in the provider module, so that the example tests use the latest example of the new logging API.

Fixes: #319 
cc @johanbrandhorst 

### DIFF
--- a/interceptors/logging/interceptors_test.go
+++ b/interceptors/logging/interceptors_test.go
@@ -95,15 +95,16 @@ func (l *mockLogger) Lines() []LogLine {
 	l.m.Lock()
 	defer l.m.Unlock()
 
-	retLines := make([]LogLine, 0, len(l.lines))
+	retLines := make([]LogLine, len(l.lines))
 	copy(retLines, l.lines)
+
 	return retLines
 }
 
 func (l *mockLogger) Log(lvl logging.Level, msg string) {
 	l.m.Lock()
 	defer l.m.Unlock()
-
+	// fmt.Printf("The fields in the Log function ---------- %v\n", l.fields)
 	line := LogLine{
 		lvl:    lvl,
 		msg:    msg,
@@ -113,20 +114,44 @@ func (l *mockLogger) Log(lvl logging.Level, msg string) {
 	for i := 0; i < len(l.fields); i += 2 {
 		line.fields[l.fields[i]] = l.fields[i+1]
 	}
+	l.lines = append(append([]LogLine{}, l.lines...), line)
 
-	l.lines = append(l.lines, line)
+	// fmt.Printf("\nAddress -- %v ", &l)
+	// fmt.Printf("Time Stamp --- %v ", time.Now().UTC())
+	// fmt.Printf(" During Logging --- MSG : %v %v\n", msg, line.fields)
+	// fmt.Printf("During Logging ----------- %v\n", l.lines)
+	// fmt.Printf("During Logging ----------- %v", l.fields)
 }
 
 func (l *mockLogger) With(fields ...string) logging.Logger {
-	logger := &mockLogger{
-		lines: l.Lines(),
-	}
-
 	l.m.Lock()
 	defer l.m.Unlock()
 
-	copy(logger.fields, l.fields)
-	logger.fields = append(logger.fields, fields...)
+	// logger := &mockLogger{}
+	logger := l
+	// logger.lines = l.lines
+	// fmt.Printf("FIELDS -------------- %v\n", fields)
+	// logger.lines = []LogLine{}
+	// logger.lines = append(logger.lines, l.lines...)
+	// logger.lines = make([]LogLine, len(l.lines))
+	// copy(logger.lines, l.lines)
+	// logger.lines = l.lines
+
+	// logger.fields = make(logging.Fields, len(l.fields))
+	// copy(logger.fields, l.fields)
+
+	// fmt.Printf("Address -- %v ", &l)
+	// fmt.Printf("Time Stamp --- %v ", time.Now())
+	// fmt.Printf("Old Lines ---------- %v\n", l.lines)
+	// fmt.Printf("Before adding --------   %v\n", logger.lines)
+	logger.fields = append(append(logging.Fields{}, l.fields...), fields...)
+	// fmt.Printf("LINES ------------- %v\n", fields)
+	// for _, elem := range logger.fields {
+	// 	fmt.Printf("New Value ----------- %v", elem)
+	// }
+	// fmt.Printf("After Adding --------- %v \n", logger.lines)
+	// fmt.Printf("New Line-----------%v============%v", logger.lines, logger.fields)
+
 	return logger
 }
 
@@ -136,8 +161,8 @@ type baseLoggingSuite struct {
 }
 
 func (s *baseLoggingSuite) SetupTest() {
-	s.logger.fields = s.logger.fields[:0]
-	s.logger.lines = s.logger.lines[:0]
+	s.logger.fields = logging.Fields{}
+	s.logger.lines = []LogLine{}
 }
 
 func customClientCodeToLevel(c codes.Code) logging.Level {
@@ -220,7 +245,9 @@ func (s *loggingClientServerSuite) TestPing() {
 	assert.Equal(s.T(), logging.DEBUG, clientFinishCallLogLine.lvl)
 	assert.Equal(s.T(), "finished call", clientFinishCallLogLine.msg)
 	clientFinishCallFields := assertStandardFields(s.T(), logging.KindClientFieldValue, clientFinishCallLogLine.fields, "Ping", interceptors.Unary)
-	clientFinishCallFields.AssertNextFieldNotEmpty(s.T(), "grpc.start_time").
+	clientFinishCallFields.AssertNextField(s.T(), "grpc.request.value", "something").
+		AssertNextFieldNotEmpty(s.T(), "peer.address").
+		AssertNextFieldNotEmpty(s.T(), "grpc.start_time").
 		AssertNextFieldNotEmpty(s.T(), "grpc.request.deadline").
 		AssertNextField(s.T(), "grpc.code", "OK").
 		AssertNextFieldNotEmpty(s.T(), "grpc.time_ms").AssertNoMoreTags(s.T())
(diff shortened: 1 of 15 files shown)


interceptors/logging/interceptors_test.go
```diff
@@ -110,11 +110,16 @@ func (l *baseMockLogger) Lines() []LogLine {
 
 type mockLogger struct {
 	*baseMockLogger
-
+	m      sync.Mutex
 	fields logging.Fields
 }
 
 func (l *mockLogger) Log(lvl logging.Level, msg string) {
+
```
johanbrandhorst, line 118: Could you remove this newline please :grin:?

interceptors/logging/interceptors_test.go
```diff
@@ -110,11 +110,16 @@ func (l *baseMockLogger) Lines() []LogLine {
 
 type mockLogger struct {
 	*baseMockLogger
-
+	m      sync.Mutex
```
johanbrandhorst, line 113: Do we need a new mutex? Can't we use the same one we're embedding with `baseMockLogger`?

johanbrandhorst: Having read this a bit more closely, I think we should just merge `baseMockLogger` and `mockLogger`. I don't see what having two different structs here gives us, and it's really confusing to read. If we have a single `mockLogger` with a mutex that should make it easy to reason about where the locks need to be.

What do you think?

johanbrandhorst (changes requested):
I think we need to rethink this `mockLogger`.

interceptors/logging/interceptors_test.go
yashrsharma44: Sure :sunglasses:

interceptors/logging/interceptors_test.go
yashrsharma44: Hmm, I guess, we should merge both the `baseMockLogger` and `mockLogger`, which doesn't make sense to break down the logger into two parts.

Regarding the locks, I read that append operations and map operations are not concurrent safe, so I was using it as a rule of thumb for using the locks there. However, if you have more suggestions about the approach, I would be happy to discuss!

johanbrandhorst: I'm confused, I looked at the PR again and it doesn't look like you've made the change to merge these two loggers. Do you agree that this is a good idea? If so, could you please try to do it in this PR?

yashrsharma44: I kind of clicked on the review option by mistake :stuck_out_tongue: I am currently working on merging the loggers, will ping you in slack once done.

yashrsharma44:
I had a few runs with the non-refactored code( the current setup) and the refactored code - 
```go
// THIS IS THE REFACTORED CODE WITH DATA RACE
type mockLogger struct {
	fields logging.Fields
	m      sync.Mutex
	lines  []LogLine
}

func (l *mockLogger) Lines() []LogLine {
	l.m.Lock()
	defer l.m.Unlock()

	return l.lines
}

func (l *mockLogger) Log(lvl logging.Level, msg string) {
	// since map isn't safe for concurrency, we use a lock
	l.m.Lock()
	defer l.m.Unlock()

	line := LogLine{
		lvl:    lvl,
		msg:    msg,
		fields: map[string]string{},
	}

	for i := 0; i < len(l.fields); i += 2 {
		line.fields[l.fields[i]] = l.fields[i+1]
	}
        l.lines = append(l.lines, line)
}

func (l *mockLogger) With(fields ...string) logging.Logger {
	// Append twice to copy slice, so we don't reuse array.
	l.m.Lock()
	defer l.m.Unlock()

	return &mockLogger{lines: l.lines, fields: append(append(logging.Fields{}, l.fields...), fields...)}
}
```
The code surprisingly doesn't have any data race for the current setup(using `baseMockLogger` and `mockLogger`) but when I merge it, there seems to be data race in spite of using locks .

I also want to **_merge the two loggers into one_**, but I am not able to find the issue of the failing tests due to the merged logger.
Suggestions are welcome.

interceptors/logging/interceptors_test.go
```diff
@@ -137,6 +130,7 @@ type baseLoggingSuite struct {
 }
 
 func (s *baseLoggingSuite) SetupTest() {
+
```
johanbrandhorst, line 133: Remove this newline please

johanbrandhorst (changes requested):
I've suggested some changes that should fix any existing race conditions in the new combined logger.

interceptors/logging/interceptors_test.go
```diff
@@ -88,47 +85,43 @@ func (l LogLines) Swap(i, j int) {
...
+	lines  []LogLine
 }
 
-func (l *baseMockLogger) Add(line LogLine) {
+func (l *mockLogger) Lines() []LogLine {
 	l.m.Lock()
 	defer l.m.Unlock()
 
-	l.lines = append(l.lines, line)
+	return l.lines
```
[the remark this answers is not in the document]
yashrsharma44: Sure

interceptors/logging/interceptors_test.go
```diff
@@ -88,47 +85,43 @@ func (l LogLines) Swap(i, j int) {
...
+	l.lines = append(l.lines, line)
 }
 
 func (l *mockLogger) With(fields ...string) logging.Logger {
 	// Append twice to copy slice, so we don't reuse array.
-	return &mockLogger{baseMockLogger: l.baseMockLogger, fields: append(append(logging.Fields{}, l.fields...), fields...)}
+	l.m.Lock()
+	defer l.m.Unlock()
+
+	return &mockLogger{lines: l.lines, fields: append(append(logging.Fields{}, l.fields...), fields...)}
```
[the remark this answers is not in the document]
yashrsharma44: Yeah sounds good!

interceptors/logging/interceptors_test.go
yashrsharma44: Yeah

interceptors/logging/interceptors_test.go
```diff
@@ -88,47 +85,43 @@ func (l LogLines) Swap(i, j int) {
...
 	l.m.Lock()
 	defer l.m.Unlock()
 
-	l.lines = append(l.lines, line)
+	return l.lines
 }
 
-func (l *baseMockLogger) Lines() []LogLine {
+func (l *mockLogger) Log(lvl logging.Level, msg string) {
+	// since map isn't safe for concurrency, we use a lock
```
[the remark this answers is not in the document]
yashrsharma44: Yup, nothing is safe :fearful:

yashrsharma44:
I applied the changes as you suggested, the data race issue has been removed, however, the fields are returning `0` data - 
```powershell
TestSuite/TestPing: interceptors_test.go:201: 
        	Error Trace:	interceptors_test.go:201
        	Error:      	"[]" should have 4 item(s), but has 0
        	Test:       	TestSuite/TestPing
=== RUN   TestSuite/TestPingError_WithCustomLevels
=== RUN   TestSuite/TestPingError_WithCustomLevels/Internal_must_remap_to_WarnLevel_in_DefaultClientCodeToLevel
    TestSuite/TestPingError_WithCustomLevels/Internal_must_remap_to_WarnLevel_in_DefaultClientCodeToLevel: interceptors_test.go:312: 
        	Error Trace:	interceptors_test.go:312
        	Error:      	"[]" should have 4 item(s), but has 0
        	Test:       	TestSuite/TestPingError_WithCustomLevels/Internal_must_remap_to_WarnLevel_in_DefaultClientCodeToLevel

```

yashrsharma44:
I read the previous code closely, and wrote some notes about the working of the code - 

Untitled-2020-10-14-1001

* We have a `baseMockerLogger` which is a struct that contains a slice `[]LogLine` that contains slice of results once we `Log` in the data. 
* We have a `mockLogger` which contains a logger with a specific field.
* By using the `With` command, it creates another `mockLogger`, and copies the previous fields into the new logger along with the fields passed in the With function

So if we want to have a single logger, we need to think how to accumulate the final results in a variable that is shared among all the mockLoggers, so that when they perform the logging using `Log` the data is appended to that shared variable.

I think the current setup is good for the logic, maybe we can change the **_name of the `baseMockLogger`, to something `sharedResults`_**, that way it is clear why we are using a shared struct for storing results.

Do you have any idea to refactor this logic, because the current setup with name changes sounds good to me?
cc @johanbrandhorst

interceptors/logging/interceptors_test.go
yashrsharma44: I think we should keep the current structure, and change the name of `baseMockLogger` to `sharedResults` or `sharedLogLines` : More Info - 
https://github.com/grpc-ecosystem/go-grpc-middleware/pull/354#issuecomment-708153888

johanbrandhorst:
Honestly I don't have enough understanding of this part of the code to know. I figured my suggested changes would fix the race, maybe we need to update how the tests read the final lines?

yashrsharma44:
Yeah seems so, if we want a single logger, we need to think of how would we compare the results of logging.

interceptors/logging/interceptors_test.go
```diff
@@ -88,47 +85,50 @@ func (l LogLines) Swap(i, j int) {
...
-	l.m.Lock()
-	defer l.m.Unlock()
-
-	return l.lines
+	return retLines
 }
 
 type mockLogger struct {
-	*baseMockLogger
+	*sharedResults
```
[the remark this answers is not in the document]
yashrsharma44: Great, thanks for confirming what I understood! 

> So we can either sort lines before comparing or remove sharedResult and keep references to all create loggers with With? Whatever works. Maybe if you name it output it will make more sense? (:

I think, I would keep the original method of comparing the results, that sounds much more dependable. The references might create a bug-prone experience for comparing the results, so I would stick with the original method.

I have renamed the `sharedResults` into `mockStdOutput`, and have added some comments for how we are using it as a shared struct. Would be happy to add more comments if needed. 

I have also sorted the output lines so that we have a definitive order for comparing the values.

yashrsharma44:
I have addressed @bwplotka comments, PTAL when you are free :sunglasses:

yashrsharma44:
PR Reviewers : It would be nice if you could add the label `hacktoberfest-accepted` **_in this_** PR, so that this PR counts for hacktoberfest :stuck_out_tongue:

johanbrandhorst:
Sorry for the radio silence on this, I haven't had the time to look at this for a while. I was hoping Bartek could get this over the line but he's been busy as well. We haven't forgotten about your work @yashrsharma44, and we appreciate it :hugs:.

yashrsharma44:
ping @bwplotka Please review this whenever you are free :hugs:

bwplotka (approved):
LGTM!

PR #358, opened 2020-10-29
util/metautils: fix double encoding bug

neena:
Previously, when a key with the suffix "-bin" was added to the metadata it would be base64 encoded twice: once in this package, and again in the GRPC over HTTP/2 implementation. However, this package does not include an automatic base64 decode when "getting" values, which meant the "set" value and the "get" value would not be the same.

Since GRPC already handles the base64 encoding, this PR removes the redundant base64 encoding here. 

Kudos to @MatthewDolan for debugging help!


neena:
@googlebot I signed it!

johanbrandhorst (approved):
This seems reasonable to me

johanbrandhorst:
Thanks for your contribution! Could you cherry pick this against v2 as well?

PR #364, opened 2020-11-24
[v2] Add skip interceptor

### LINKED ISSUE #363: Add skip package to prevent specific methods affect by middlewares
Middlewares impact all methods once they apply, but not all methods are likely to be influenced. For instance, login methods and reflection methods should not affect by the `auth` middleware; users might not like to log or trace access records of reflection methods.

Instead of adding this feature to certain middleware, we should add an independent package, which might be called the `skip` package, to handle this issue once for all.

Here is the possible implementation:

```go
package grpc_skip

type UnaryFilter func(info *grpc.UnaryServerInfo) bool
type StreamFilter func(info *grpc.StreamServerInfo) bool

func UnaryServerInterceptor(in grpc.UnaryServerInterceptor, filter UnaryFilter) grpc.UnaryServerInterceptor {}
func StreamServerInterceptor(in grpc.StreamServerInterceptor, filter StreamFilter) grpc.StreamServerInterceptor {}
```

Examples:
```go
server := grpc.NewServer(
		grpc.UnaryInterceptor(grpc_middleware.ChainUnaryServer(
			grpc_ctxtags.UnaryServerInterceptor(),
			grpc_skip.UnaryServerInterceptor(
				grpc_zap.UnaryServerInterceptor(logger),
				unaryFilter,
			),
			grpc_skip.UnaryServerInterceptor(
				grpc_auth.UnaryServerInterceptor(authenticator.AuthFunc),
				unaryFilter,
			),
		))
	)
```

XSAM:
Resolve #363 

It's hard to transfer a `interceptors.UnaryServerInterceptor` into `interceptors.Reporter`, so I use the old way to implement this feature.

### DIFF
--- a/interceptors/skip/examples_test.go
+++ b/interceptors/skip/examples_test.go
@@ -5,16 +5,28 @@ import (
 
 	"google.golang.org/grpc"
 
+	middleware "github.com/grpc-ecosystem/go-grpc-middleware/v2"
 	"github.com/grpc-ecosystem/go-grpc-middleware/v2/auth"
 	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors"
 	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/skip"
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/tags"
 )
 
 // Simple example of skipping auth interceptor in the reflection method.
 func Example_initialization() {
 	_ = grpc.NewServer(
-		grpc.UnaryInterceptor(skip.UnaryServerInterceptor(auth.UnaryServerInterceptor(exampleAuthFunc), ReflectionFilter)),
-		grpc.StreamInterceptor(skip.StreamServerInterceptor(auth.StreamServerInterceptor(exampleAuthFunc), ReflectionFilter)),
+		grpc.UnaryInterceptor(skip.UnaryServerInterceptor(auth.UnaryServerInterceptor(dummyAuth), SkipReflectionService)),
+		grpc.StreamInterceptor(skip.StreamServerInterceptor(auth.StreamServerInterceptor(dummyAuth), SkipReflectionService)),
+	)
+}
+
+func Example_chain() {
+	_ = grpc.NewServer(
+		grpc.UnaryInterceptor(skip.UnaryServerInterceptor(
+			middleware.ChainUnaryServer(
+				tags.UnaryServerInterceptor(),
+				auth.UnaryServerInterceptor(dummyAuth),
+			), SkipReflectionService)),
 	)
 }
 
@@ -22,6 +34,6 @@ func dummyAuth(ctx context.Context) (context.Context, error) {
 	return ctx, nil
 }
 
-func SkilReflectionService(ctx context.Context, gRPCType interceptors.GRPCType, service string, method string) bool {
-	return service == "grpc.reflection.v1alpha.ServerReflection"
+func SkipReflectionService(ctx context.Context, gRPCType interceptors.GRPCType, service string, method string) bool {
+	return service != "grpc.reflection.v1alpha.ServerReflection"
 }
--- a/interceptors/skip/examples_test.go
+++ b/interceptors/skip/examples_test.go
@@ -18,10 +18,10 @@ func Example_initialization() {
 	)
 }
 
-func exampleAuthFunc(ctx context.Context) (context.Context, error) {
+func dummyAuth(ctx context.Context) (context.Context, error) {
 	return ctx, nil
 }
 
-func ReflectionFilter(ctx context.Context, gRPCType interceptors.GRPCType, service string, method string) bool {
+func SkilReflectionService(ctx context.Context, gRPCType interceptors.GRPCType, service string, method string) bool {
 	return service == "grpc.reflection.v1alpha.ServerReflection"
 }
--- a/interceptors/skip/examples_test.go
+++ b/interceptors/skip/examples_test.go
@@ -1,0 +1,27 @@
+package skip_test
+
+import (
+	"context"
+
+	"google.golang.org/grpc"
+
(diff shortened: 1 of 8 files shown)


interceptors/skip/examples_test.go
```diff
@@ -0,0 +1,27 @@
...
+		grpc.UnaryInterceptor(skip.UnaryServerInterceptor(auth.UnaryServerInterceptor(exampleAuthFunc), ReflectionFilter)),
+		grpc.StreamInterceptor(skip.StreamServerInterceptor(auth.StreamServerInterceptor(exampleAuthFunc), ReflectionFilter)),
+	)
+}
+
+func exampleAuthFunc(ctx context.Context) (context.Context, error) {
+	return ctx, nil
+}
+
+func ReflectionFilter(ctx context.Context, gRPCType interceptors.GRPCType, service string, method string) bool {
```
bwplotka, line 25:
```suggestion
func SkilReflectionService(ctx context.Context, gRPCType interceptors.GRPCType, service string, method string) bool {
```

interceptors/skip/examples_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+// Simple example of skipping auth interceptor in the reflection method.
+func Example_initialization() {
+	_ = grpc.NewServer(
+		grpc.UnaryInterceptor(skip.UnaryServerInterceptor(auth.UnaryServerInterceptor(exampleAuthFunc), ReflectionFilter)),
+		grpc.StreamInterceptor(skip.StreamServerInterceptor(auth.StreamServerInterceptor(exampleAuthFunc), ReflectionFilter)),
+	)
+}
+
+func exampleAuthFunc(ctx context.Context) (context.Context, error) {
```
bwplotka, line 21:
```suggestion
func dummyAuth(ctx context.Context) (context.Context, error) {
```

bwplotka: It's function so no need for `Func` suffix (:

interceptors/skip/interceptor.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"context"
+
+	"google.golang.org/grpc"
+
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors"
+)
+
+type Filter func(ctx context.Context, gRPCType interceptors.GRPCType, service string, method string) bool
```
bwplotka, line 11:
```suggestion
type Filter func(ctx context.Context, typ interceptors.GRPCType, svc string, method string) bool
```

bwplotka: let's make it conceise

interceptors/skip/interceptor.go
```diff
@@ -0,0 +1,43 @@
...
+// UnaryServerInterceptor returns a new unary server interceptor that determines whether to skip the input interceptor.
+func UnaryServerInterceptor(in grpc.UnaryServerInterceptor, filter Filter) grpc.UnaryServerInterceptor {
+	if filter == nil {
+		return in
+	}
+
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		service, method := interceptors.SplitMethodName(info.FullMethod)
+		if filter(ctx, interceptors.Unary, service, method) {
+			// Skip interceptor
```
bwplotka, line 22:
```suggestion
			// Skip interceptor.
```

interceptors/skip/interceptor.go
```diff
@@ -0,0 +1,43 @@
...
+// StreamServerInterceptor returns a new streaming server interceptor that determines whether to skip the input interceptor.
+func StreamServerInterceptor(in grpc.StreamServerInterceptor, filter Filter) grpc.StreamServerInterceptor {
+	if filter == nil {
+		return in
+	}
+
+	return func(srv interface{}, ss grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
+		service, method := interceptors.SplitMethodName(info.FullMethod)
+		if filter(ss.Context(), interceptors.StreamRPCType(info), service, method) {
+			// Skip interceptor
```
bwplotka, line 38:
```suggestion
			// Skip interceptor.
```

bwplotka (approved):
Nice, I like it, but curious why denylist was chosen instead of allowlist approach? :thinking: 

To me allowlist approach might be more comprehensive. E.g If you want to use logging middleware only on service X. I will add `SKIP -> Logging, filter service != x` Doable but quite big inversion (: 

I think we could consider this, at the end it's flexible enough anyway, just curious what is easier to use.

Skip might be nice for auth aspect, so by default you deny from auth etc.. Maybe does not matter that much :hugs: 

LGTM, overall :muscle:

XSAM:
> To me allowlist approach might be more comprehensive.

I agree with that. At the first design, `If service == x then skip` seems good, but the function is called `filter`. Therefore, `filter service != x` seems better.

XSAM:
@johanbrandhorst Shall we merge it?

johanbrandhorst (changes requested):
LGTM, but could we add this new interceptor to the README too please? Sorry for the delayed review.

XSAM:
@googlebot I consent.

XSAM:
@johanbrandhorst It seems like the CLA bot is not working.

johanbrandhorst:
I think @bwplotka might need to consent too.

johanbrandhorst (approved):

XSAM:
@bwplotka Shall we?

yashrsharma44:
Yeah, seems like @bwplotka also need to consent 😛

bwplotka (approved):
Sorry for lag, if you think it makes sense I am happy as well. LGTM!

bwplotka:
@googlebot I consent.

bwplotka:
All is good to me, so merging.

PR #366, opened 2020-11-27
Fix zap grpclogger verbosity check

bouk:
Currently it only logs entries that are *lower* or equal to the
configured level, when you want the opposite. This means that by default
only INFO messages are logged, and anything more urgent is dropped.

### DIFF
--- a/logging/zap/grpclogger.go
+++ b/logging/zap/grpclogger.go
@@ -114,5 +114,5 @@ func (l *zapGrpcLoggerV2) Fatalf(format string, args ...interface{}) {
 }
 
 func (l *zapGrpcLoggerV2) V(level int) bool {
-	return level <= l.verbosity
+	return l.verbosity <= level
 }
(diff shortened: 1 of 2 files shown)


johanbrandhorst:
Hi, thanks for this PR! Could you add a test to check that this is indeed the correct behavior? Thanks!

bouk:
Added!

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution! Could you please cherry pick this against the v2 branch?

bouk:
The v2 branch doesn't seem to do anything with the global grpc logger

fxrlv:
verbosity **IS NOT** about logging level
`grpclog` package logs error by calling method `(*zapGrpcLoggerV2).Error*`
verbosity controls how much details we log, **NOT** level
could you please revert this pull request?

Issue #150, opened 2018-07-13
AuthFunc doesn't contain the FullMethod info


thincal:
Hi,

If I want to use the `AuthFunc` to do the authorization *centrally*, it needs to know which RPC method invoked.

I know that `AuthFuncOverride` could be used to do it per service, but I have a central authorization module to do this.

So what can I do ?

Thanks,
Song

ZachEddy:
@domgreen I ran in to a similar issue. Basically, the request `context` inside my `AuthFunc` doesn't provide enough information to authorize a given request. I would really like to have access to the request itself inside the `AuthFunc` I wrote.

To give you some background, my gRPC function looks something like this.
```
// GetPetByID returns information about a given pet in the pet store
func (s server) GetPetByID(ctx context.Context, req *pb. GetPetByIDRequest) (*pb. GetPetByIDResponse, error)
```
The `req` parameter contains an `id` field that uniquely identifies a pet in the database. With this in mind, my authorization logic is, _"can user XYZ allowed to get information about this pet ABC?"_

It turns out that `id` would be useful to have, but it's not inside the context.

Here are some general thoughts about how I should proceed. I would be interested to hear your input. 

- In addition to middleware, I should make an authorization check inside the service itself. The service implementation will have access to the `id` field (seems reasonable, but requires two round-trip calls to our central authorization service).

- Write my own authorization middleware that is basically identical to this repository, but `AuthFunc` will take `req` as an additional argument (not ideal, but I can import the `grpc_auth` package and use its helpers).

- Open a pull request to `go-grpc-middleware` that adds `req` to the `AuthFunc` type (break code for lots of people).

seinshah:
@ZachEddy I also have the same requirement which would make the code more clear if I perform the complete authentication before calling the method. Since, some scenario need `req` and some need `info`, I'd suggest the following approach:

```go
type AuthFuncExtended func(ctx context.Context, req interface{}, fullMethodName string) (context.Context, error)

type ServiceAuthFuncOverrideExtended interface {
	AuthFuncOverride(ctx context.Context, req interface{}, fullMethodName string) (context.Context, error)
}

func UnaryServerInterceptorExtended(authFunc AuthFuncExtended) grpc.UnaryServerInterceptor {
	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
		var newCtx context.Context
		var err error
[...]

PR #374, opened 2020-12-18
Add authorization middleware

jacobweinstock:
Addresses https://github.com/grpc-ecosystem/go-grpc-middleware/issues/150
The existing auth middleware is catered to authentication. This creates a new auth middleware that can be used specifically for authorization as it takes in the full method name. The tokens scopes/claims can then be checked against allowed scopes to full method names.


jacobweinstock:
> @googlebot I signed it!

@googlebot I signed it!

johanbrandhorst:
Hi Jacob. I appreciate this thorough contribution, but shouldn't this already be possible to support in the existing AuthFunc using https://pkg.go.dev/google.golang.org/grpc#Method to get the method from the context?

jacobweinstock:
Oh man, you are right! Apologies, closing this.

PR #379, opened 2021-01-14
Add OpenMetrics(Prometheus) in the provider module

### LINKED ISSUE #346: v2: Move https://github.com/grpc-ecosystem/go-grpc-prometheus as another provider. Deprecate repo.
This was discussed initially with @brancz https://github.com/grpc-ecosystem/go-grpc-middleware/issues/275#issuecomment-593261947

I think the decision is to incorporate this Prometheus middleware here. 

Not a blocker for first v2 release.

yashrsharma44:
Fixes #346 

Some questions:
> - [x] Should we integrate Prometheus in the metrics package, or provide it as another provider?

Have Prometheus as a separate provider, which follows Open Metrics convention, we need to decide what do we want to name the package though since the repo is Open Metrics compliant

- [x] How should we add in options for enabling the histograms for different purposes?

This PR is a WIP PR, **_however suggestions regarding the implementation details are welcome because most of the structure of the code is final_**. The interface for the reporter has been tweaked a little for enabling the Timer feature from Prometheus.

cc @bwplotka @brancz

TODO:
- [x] Write tests for the new provider
- [ ] Write docs for the same


bwplotka (commented):
Amazing work! I think it's literally what we need in the first iteration, but before full `v2` let's address all pain points @brancz suggested in https://gophers.slack.com/archives/CNJL30P4P/p1610112810036200?thread_ts=1610112652.036000&cid=CNJL30P4P 

You first question should be probably 2 questions to answer:

1. Do we want Prometheus as core vs provider?

I don't think we can put it in core, because Prometheus client might be not what user will use and what if user don't want to use metric interceptor at all?  Then having in core will dictate them to have Prometheus dependency in code.

(Discussed in Thanos Contributor Hours)

2. IF we want Prometheus as a provider do we need any "metrics" interceptor in the core.

We have some similarities... like recording specific things.

There is a good rule which is YAGNI - unless there is already 3 or more specific "users" of the vendor-agnostic metric interceptor, there is no point creating one, because we don't know even if there will be any other "user" than Prometheus, and we don't know what they want.

(Discussed in Thanos Contributor Hours)

3 .Dario: Maybe open telemetry is a common thing that allows bridge.

My understanding is that the current state is that Open Telemetry does not support Open Metrics or actually anything without extra bridges (sidecars, agent), so open metrics via the most popular Prometheus client sounds like a good choice. Since it will separate module, this project will work with open telemetry just fine (: (in or out of this repo)

providers/prometheus/go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/grpc-ecosystem/go-grpc-middleware/providers/prometheus
+
+go 1.15
```
bwplotka, line 3: I think we agreed we put all of this into `openmetrics` directory no? Why it is in Prometheus now?

providers/prometheus/go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/grpc-ecosystem/go-grpc-middleware/providers/prometheus
```
bwplotka, line 1: Since we want to release all modules once, we need v2 here.
```suggestion
module github.com/grpc-ecosystem/go-grpc-middleware/providers/openmetrics/v2
```

providers/prometheus/interceptors.go
```diff
@@ -0,0 +1,133 @@
...
+type reporter struct {
+	clientMetrics   *ClientMetrics
+	serverMetrics   *ServerMetrics
+	typ             interceptors.GRPCType
+	service, method string
+	startTime       time.Time
+	kind            Kind
+}
+
+func (r *reporter) StartTimeCall(startTime time.Time, callType string) interceptors.Timer {
```
bwplotka, line 22: I think I actually remove this as part of core reporter interface as we can totally put this logic into `reporter` function (below) logic, no? 

Look on `logging` interceptor it has exactly the same mechanism to report latency. (:

bwplotka (changes requested):
Nice! Going in good direction but still a few major things:

1. We agreed on `openmetrics` I think?
2. I removed initially `StartTimeCall` reporter method for a reason (: We can have much simpler code if (less methods in in terface = better!). See comments below.

Thanks for helping 🤗

providers/prometheus/go.mod
yashrsharma44: Thanks, I was confused with the naming, changing to openmetrics!

yashrsharma44: Done

providers/prometheus/go.mod
yashrsharma44: Okay, changed!

providers/prometheus/interceptors.go
yashrsharma44: Okay, I now get your point. Basically use the duration parameter of the reporter functions to get the job done. Nice optimization suggestion!

providers/openmetrics/client_metrics.go
```diff
@@ -0,0 +1,168 @@
...
+		m.clientStreamRecvHistogram.Collect(ch)
+	}
+	if m.clientStreamSendHistogramEnabled {
+		m.clientStreamSendHistogram.Collect(ch)
+	}
+}
+
+// EnableClientHandlingTimeHistogram turns on recording of handling time of RPCs.
+// Histogram metrics can be very expensive for Prometheus to retain and query.
+func (m *ClientMetrics) EnableClientHandlingTimeHistogram(opts ...HistogramOption) {
```
yashrsharma44, line 123: @bwplotka How should we enable the options? Much similar to logging middlewares, or do you have something different in mind? 🤔

providers/openmetrics/interceptors.go
```diff
@@ -0,0 +1,130 @@
...
+			r.receiveTimer = prometheus.NewTimer(hist)
+		}
+	case KindServer:
+		if r.serverMetrics.serverHandledHistogramEnabled {
+			r.startTime = time.Now()
+		}
+		r.serverMetrics.serverStartedCounter.WithLabelValues(string(r.typ), r.service, r.method).Inc()
+	}
+
+	// TODO: @yashrsharma44 - What should we use instead of the context.Background()?
```
yashrsharma44, line 112: What should we return instead of the background context? What should we return instead?

providers/openmetrics/options.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	prom "github.com/prometheus/client_golang/prometheus"
+	"google.golang.org/grpc/status"
+)
+
+// Helper methods
+func FromError(err error) (s *status.Status, ok bool) {
+	return status.FromError(err)
+
+	// TODO: @yashrsharma44 - discuss if we require more error handling from the previous package
```
yashrsharma44, line 12: In the grpc prometheus, they had more complex handling of error [here](https://github.com/grpc-ecosystem/go-grpc-prometheus/blob/9abf3eb82b4a313b1a28f370a7ef8ffe0667c681/packages/grpcstatus/grpcstatus.go#L37). Do we want to have this logic here?

yashrsharma44:
I haven't tested the middleware locally, so I will have more questions when done 😛

bwplotka:
Can we make it not a draft when ready for review? (:

yashrsharma44:
I haven't added the options, so technically isn't completed, will make it ready for review once I add the options.

providers/openmetrics/client_metrics.go
yashrsharma44: I have enabled the previous convention to enable histogram options. Let me know if the changes make sense

yashrsharma44:
@bwplotka 
I have added the options for enabling the histogram. PTAL whenever you are free 🤗

prmsrswt (commented):
Overall LGTM, just a few comments. :)

interceptors/reporter.go
```diff
@@ -14,6 +14,23 @@ import (
 
 type GRPCType string
 
+// timer is a helper interface to time functions.
```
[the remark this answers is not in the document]
yashrsharma44: Done!

interceptors/reporter.go
```diff
@@ -14,6 +14,23 @@ import (
...
+// Useful for interceptors to record the total
+// time elapsed since completion of a call.
+type Timer interface {
+	ObserveDuration() time.Duration
+}
+
+// NoOp timer
```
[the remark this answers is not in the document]
yashrsharma44: Cool.

providers/openmetrics/client_interceptor.go
```diff
@@ -0,0 +1,59 @@
...
+	return interceptors.StreamClientInterceptor(&reportable{})
+}
+
+// EnableClientHandlingTimeHistogram turns on recording of handling time of
+// RPCs. Histogram metrics can be very expensive for Prometheus to retain and
+// query. This function acts on the DefaultClientMetrics variable and the
+// default Prometheus metrics registry.
+func EnableClientHandlingTimeHistogram(opts ...HistogramOption) {
+	DefaultClientMetrics.EnableClientHandlingTimeHistogram(opts...)
+	openmetrics.Register(DefaultClientMetrics.clientHandledHistogram)
```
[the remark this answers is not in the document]
yashrsharma44: I think we should use `openmetrics.MustRegister` because the above function - `EnableClientHandlingTimeHistogram` is supposed to be called during server initialization, and not while scraping metrics. 

Thanks for pointing out this 😅 , would have been difficult to debug why this feature isn't working in case of error.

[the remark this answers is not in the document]
yashrsharma44: ping @bwplotka for suggestions, although using `MustRegister` sounds fair to me :).

yashrsharma44:
I have added the tests for open metrics. The lint makes some suggestions, let me know if I need to address them.

PTAL whenever free @johanbrandhorst @bwplotka 🎉

yashrsharma44:
Let's leave the examples to be added in the next PR, would be nice to separate this from this PR.

providers/openmetrics/go.mod
```diff
@@ -0,0 +1,13 @@
+module github.com/grpc-ecosystem/go-grpc-middleware/providers/openmetrics/v2
+
+go 1.15
+
+require (
+	github.com/golang/protobuf v1.4.3
+	github.com/grpc-ecosystem/go-grpc-middleware/v2 v2.0.0-rc.2
```
bwplotka, line 7: Let's update this to commit that does have your change ;p 

See similar discussion here: https://cloud-native.slack.com/archives/CL25937SP/p1611245939002300

bwplotka (approved):
LGTM, but we need to fix module to point to newest github.com/grpc-ecosystem/go-grpc-middleware/v2 version (somehow). There many different ways, but build has to pass (:

providers/openmetrics/client_interceptor.go
```diff
@@ -0,0 +1,62 @@
...
+	"google.golang.org/grpc"
+
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors"
+)
+
+var (
+	// DefaultClientMetrics is the default instance of ClientMetrics. It is
+	// intended to be used in conjunction the default Prometheus metrics
+	// registry.
+	DefaultClientMetrics = NewClientMetrics()
```
ash2k, line 14: It would be really nice to avoid adding any global/default things to the v2 of the library! Why not let users decide how to use it? If they use the global registry it can be ~1 line of code to register all metrics with it but if they don't then there is simply no way to opt out. Also, having global as default encourages everybody to use the global metrics but if you have a use case where globals don't work for you, you are having a really hard time because everybody is using them.

For example, currently there is `NewClientMetrics()` so supposedly you can create a new instance and don't use globals, but... there is not way to register metrics inside of the `ClientMetrics` with a custom registry.

Another example - what if you have several clients and some of them should have certain histograms and some of them - should not, or should have a different set of histograms? With globals used everywhere it's impossible to have distinct configuration. I think all `Enable...()` functions should be turned into options for `NewClientMetrics()` to make it all configurable per-instance.

Note that globals can always be added later but cannot be taken out once added (API breakage) so v2 is a perfect time to remove them from the API. Instead `ClientMetrics.Register(prometheus.Registerer)` can be added to make it trivial, whether you use globals or not, to register their metrics with the desired (global or not) registry. Like so:

```go
package my_global_metrics

import (
    "github.com/prometheus/client_golang/prometheus"
    "github.com/grpc-ecosystem/go-grpc-middleware/providers/openmetrics"
)

var (
    DefaultClientMetrics = openmetrics.NewClientMetrics()
)
func init() {
    DefaultClientMetrics.Register(prometheus.DefaultRegisterer).
}
```

interceptors/reporter.go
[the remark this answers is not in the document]
yashrsharma44: Strange, I resolved it before, doing it right away!

interceptors/reporter.go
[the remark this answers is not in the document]
yashrsharma44: eee:)

interceptors/reporter.go
```diff
@@ -14,6 +14,23 @@ import (
...
+type noOpTimer struct {
```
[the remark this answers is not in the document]
yashrsharma44: Yeah, bad choice of naming :)

providers/openmetrics/go.mod
yashrsharma44: I think I need to add the interceptor change in a different PR, and then tag that commit.

bwplotka:
That's one option. You can use replace comment to pin to commit from your
own fork as well.
Kind Regards,
Bartek Płotka (@bwplotka)

providers/openmetrics/client_interceptor.go
bwplotka: Yes. That's definitely a plan to get rid of, but we think to do it in another PR. 

However maybe it's easy to do it now?

bwplotka: @yashrsharma44 fixing now or in next PR? (: 


I will be available whole day tomorrow so we can quickly iterate.

providers/openmetrics/go.mod
bwplotka: Still you need to put newest commit here not v2.0.0-rc.2

bwplotka (approved):

providers/openmetrics/client_interceptor.go
yashrsharma44: Hey, I will push the changes in this PR itself, by today. Sounds good?

bwplotka: Let us know!

yashrsharma44:
Current Update:
Thanks to @ash2k for reviewing and asking to remove the global state.
I have written the current setup which uses registry as a variable, and other functions associated with it.
```go
DefaultClientMetrics = openmetrics.NewClientMetrics(prometheus.DefaultRegisterer)
if err := DefaultClientMetrics.EnableClientStreamSendTimeHistogram; err != nil {
   // Handle Error
}
```
* All the options for enabling the histogram is connected with the client metrics instance, while for using it in the interceptor, pass the registry as a variable.
* `UnaryClientInterceptor` takes in registry as a variable, because I didn't wanted to integrate the metrics object with reporter. So taking the registry as a variable.

providers/openmetrics/go.mod
yashrsharma44: Yeah forgot it before, have added it now 😁

providers/openmetrics/client_interceptor.go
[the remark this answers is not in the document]
yashrsharma44: Let's propagate the error to the user because the options might be used after the registry. So let the user handle the error, rather than panicking.

providers/openmetrics/client_interceptor.go
```diff
@@ -0,0 +1,36 @@
+package metrics
+
+import (
+	openmetrics "github.com/prometheus/client_golang/prometheus"
+	"google.golang.org/grpc"
+
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors"
+)
+
+var (
```
bwplotka, line 10: Can we remove this? This is global we wanted to remove, no?

yashrsharma44: Yeah, I didn't understand it before, removing it totally.

providers/openmetrics/client_metrics.go
```diff
@@ -0,0 +1,187 @@
...
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	openmetrics "github.com/prometheus/client_golang/prometheus"
+)
+
+// ClientMetrics represents a collection of metrics to be registered on a
+// Prometheus metrics registry for a gRPC client.
+type ClientMetrics struct {
+	clientRegister openmetrics.Registerer
```
bwplotka, line 11: We can remove this (:

providers/openmetrics/client_metrics.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+// NewClientMetrics returns a ClientMetrics object. Use a new instance of
+// ClientMetrics when not using the default Prometheus metrics registry, for
+// example when wanting to control which metrics are added to a registry as
+// opposed to automatically adding metrics via init functions.
+func NewClientMetrics(clientRegistry prometheus.Registerer, counterOpts ...CounterOption) *ClientMetrics {
+	opts := counterOptions(counterOpts)
+	return &ClientMetrics{
+		clientRegister: clientRegistry,
```
bwplotka, line 38: This pattern can be further simplified.

```suggestion
func NewClientMetrics(r prometheus.Registerer, counterOpts ...CounterOption) *ClientMetrics {
	opts := counterOptions(counterOpts)
	c := &ClientMetrics{
		
		
	r.MustRegister(
	     c.clientStartedCounter,
	     ....
	     ...
	 	
```

bwplotka (commented):
Amazing, one suggestion (:

providers/openmetrics/client_metrics.go
yashrsharma44: We are using `clientRegister` in `Enable...` functions. Should I modify the `Enable..` functions to accept clientRegister?

providers/openmetrics/client_metrics.go
yashrsharma44: Sounds fair 👍

providers/openmetrics/client_metrics.go
bwplotka: Hmmm
need to look on this, happy to propose something on top of your PR (:

bwplotka: Ideally register is only needed on start!

bwplotka:
Actually you know what. Let's merge. I think we are trying to do exactly what we wanted to avoid -> adding thousands of improvements to porting PR (: 🤗 

Thanks @yashrsharma44 for your work

I will try to apply some fixes for registration on next PR if that's ok!

ti:
@yashrsharma44 @bwplotka 
There are two Fatal bug in This mr：

1. Lost server context when client call server with: metadata.NewOutgoingContext(ctx, md)
2. Metrics counter does not work

Please see add "metadata.NewOutgoingContext(" test case, you will see.

refer: https://github.com/grpc-ecosystem/go-grpc-middleware/blob/215af813c95e0f11477e603edf25f6da4f02d413/providers/openmetrics/reporter.go#L113

yashrsharma44:
> Metrics counter does not work

All the prometheus metrics, or any specific one?

>Lost server context when client call server with: metadata.NewOutgoingContext(ctx, md)

Thanks for reporting this, would be nice if you could propose a fix for this. I can add the fix in another PR.

PR #397, opened 2021-02-11
use status.Error instead of status.Errorf

### LINKED ISSUE #396: passing `err.Error()` to the format string of `grpc.Errorf` allows format string injection
https://github.com/grpc-ecosystem/go-grpc-middleware/blob/fab13c27ec1d74ec053740349a01efaae87e5aa4/validator/validator.go#L25

I think the call to `grpc.Errorf` should look something like `grpc.Errorf(codes.InvalidArgument, "%s", err.Error())`. Alternately, you could call `status.Error` instead since `grpc.Errorf` is a wrapper around `status.Errorf`.

petermattis:
Use `status.Error` instead of `status.Errorf` when the format string is
non-constant and not actually a format string. In the case of the
validator middleware, the error being supplied as a format string could
potentially contain data supplied by an attacker allowing for format
string injection. This doesn't appear to be an actual problem due to
`fmt` being safe in this regards, but it certainly isn't good practice
to provide a format string that an attacker can control.

Fixes #396

### DIFF
--- a/retry/retry.go
+++ b/retry/retry.go
@@ -236,7 +236,9 @@ func (s *serverStreamingRetryingStream) receiveMsgAndIndicateRetry(m interface{}
 	return isRetriable(err, s.callOpts), err
 }
 
-func (s *serverStreamingRetryingStream) reestablishStreamAndResendBuffer(callCtx context.Context) (grpc.ClientStream, error) {
+func (s *serverStreamingRetryingStream) reestablishStreamAndResendBuffer(
+	callCtx context.Context,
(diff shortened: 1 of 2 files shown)


johanbrandhorst (approved):
LGTM

johanbrandhorst:
Thanks for your contribution! Could you please cherry pick this fix against the v2 branch?

petermattis:
> Thanks for your contribution! Could you please cherry pick this fix against the v2 branch?

Not any time soon unfortunately.

PR #398, opened 2021-02-12
Support customization of timestamp format

### LINKED ISSUE #131: grpc.request.deadline and grpc.start_time precision is in seconds
`grpc.request.deadline` and `grpc.start_time` use `d.Format(time.RFC3339)` which means maximum precision is in seconds. I believe it would be useful to use `d.Format(time.RFC3339Nano)` at least.

Best would be a configuration option for me to format all logged timestamps as desired.

I can work on a PR for this change if it makes sense.

stanhu:
By default, RFC3339 timestamps are used, but our application uses a
custom format. This commit enables us to set the format in a consistent
manner.

Closes #131

### DIFF
--- a/logging/kit/options.go
+++ b/logging/kit/options.go
@@ -11,17 +11,19 @@ import (
 
 var (
 	defaultOptions = &options{
-		shouldLog:    grpc_logging.DefaultDeciderMethod,
-		codeFunc:     grpc_logging.DefaultErrorToCode,
-		durationFunc: DefaultDurationToField,
+		shouldLog:       grpc_logging.DefaultDeciderMethod,
+		codeFunc:        grpc_logging.DefaultErrorToCode,
+		durationFunc:    DefaultDurationToField,
+		timestampFormat: time.RFC3339,
 	}
 )
 
 type options struct {
-	levelFunc    CodeToLevel
-	shouldLog    grpc_logging.Decider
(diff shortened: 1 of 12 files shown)


johanbrandhorst (approved):
LGTM

johanbrandhorst:
Thanks for your contribution! Could you please cherry pick this against v2? It may need significant refactoring but it'd be good to have in the new version too.

stanhu:
@johanbrandhorst Sure, I took a look at the v2 branch. It looks like the new payload interceptors have a bit different signature from `logging.interceptors`:

https://github.com/grpc-ecosystem/go-grpc-middleware/blob/c1f1e53b80a7205e622f9d6c3325021d8a99ebe9/interceptors/logging/payload.go#L126-L148

It doesn't seem like we can reuse `logging.Options` here because the `Decider` and `ClientPayloadLoggingDecider` signatures are different here. A few options:

1. Introduce a `PayloadOptions`
2. Tack on the format to the end of each function
3. Expand the `Logging` interface to set timestamp formatting optoins

johanbrandhorst:
CC @bwplotka

stanhu:
I did the simplest thing with option 2 in https://github.com/grpc-ecosystem/go-grpc-middleware/pull/399, but I don't love it. Feel free to refactor as necessary.

PR #403, opened 2021-03-24
Allow extracting a sugared logger directly

marcusirgens:
When using Zap, it is common to switch between using `zap.Logger` and `zap.SugaredLogger`, so to avoid `ctxzap.Extract(ctx).Sugar()` being used all over, `ctxzap.ExtractS(ctx)` can do the same, saving 7(!) characters.


marcusirgens:
@googlebot I signed it!

johanbrandhorst:
Hi Marcus, while I appreciate the desire to contribute back, I don't think this adds enough utility to users of this library for use to include it in our API. It's easy enough for users to write this helper themselves, and adding API always increases maintenance burden, which is already something we're struggling with at the moment. I'm leaning against merging this right now, do you have a good argument for why we should still consider it?

marcusirgens:
> It's easy enough for users to write this helper themselves

Fair enough. I would also be against adding `ctxzap.Debug(ctx, ...args)` and other direct access methods to the `zap` API. The argument for is that you're essentially dealing with two "separate" loggers when using zap, one "fast" and one "slow". Accessing them directly without doing the conversion is a (small) QoL improvement, especially considering that one of the perks of the "Sugared" logger is that is less verbose ;) 

As for maintainability, the zap API is quite stable and I do not expect to see any changes anytime soon. The alternative to this is a per-user wrapper for ctxzap package, which moves the maintenance burden from your end to N users. :)

As always, thanks for responding quickly, @johanbrandhorst 👍🏻

johanbrandhorst:
You're right of course that this is realistically not adding a lot of maintenance burden, but given that we've come this far without it, I'd also argue it's not that common of an ask. If someone else adds this feature request, we'll reconsider it, for now, I'm going to close this PR. As always, thanks for your contributions 🙂.

Issue #356, opened 2020-10-07
retry: add onRetry callback


ericwenn:
Currently (v1.2.2) the retry interceptor logs retries to the trace, see https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/retry/retry.go#L49

It would be great if we could hook in our own logger here via options. Something along the lines of:
```go 
type OnRetryCallback func(ctx context.Context, attempt uint, err error)

func WithOnRetryCallback(fn OnRetryCallback) CallOption
```

**Additional context**
We have our own request logger for gRPC Client calls, which we are currently wrapping outside `retry.UnaryClientInterceptor`, to only log the "final" response (after potential retries). We only retry `Unavailable` status codes. This is great as we are not alerted for any intermittent `Unavailable` status codes, but we have come across situations where we think a request should have been retried but can't really know what the retry interceptor is doing.

** Maintainer edit**: merged this with #375 and updated description.

johanbrandhorst:
This seems reasonable, but would you be willing to add it to the v2 branch instead? We're moving away from adding new features to master.

ericwenn:
I see, completely missed there was a separate v2 track going on with this repository, my bad.
Just for understanding, would this mean the timeline of this feature (should it happen) would be tied to the release of v2 of this library?

johanbrandhorst:
Yes. We're working on v2 right now, you can track the progress on the issue tracker.

ericwenn:
v2 is fine by me, do you need any action from me?

johanbrandhorst:
We'd love to have you implement this feature on the v2 branch :). We're probably not going to get around to it ourselves, we're pretty contribution-driven. You've already found the place to add it I believe so if you can put up a PR with a test that ought to be enough to have this merged :smile:.

ericwenn:
Gotcha, will look into it! thanks for swift replies.

johanbrandhorst:
I merged #375 into this, I think this is still a good proposal, but we can make it more generic than just for logging. The signatures stay the same but we'd call it something more generic.

flapjack103:
Thanks for this addition! I recently added the new call option `WithOnRetryCallback` to be passed to the UnaryClientInterceptor and request errors are invoking the callback function but what I'm seeing is that this is invoked even for non-retriable status codes. This contradicts the [documentation](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/v2/interceptors/retry/options.go#L51).

[...]

PR #405, opened 2021-04-02
add onRetryCallback callback function

shamil:
Implements the suggested solutoin in https://github.com/grpc-ecosystem/go-grpc-middleware/issues/356

### DIFF
--- a/interceptors/retry/options.go
+++ b/interceptors/retry/options.go
@@ -27,7 +27,7 @@ var (
 			return BackoffLinearWithJitter(50*time.Millisecond /*jitter*/, 0.10)(attempt)
 		}),
 		onRetryCallback: OnRetryCallback(func(ctx context.Context, attempt uint, err error) {
-			// By default we don't have any callback logic to execute
+			logTrace(ctx, "grpc_retry attempt: %d, backoff for %v", attempt, err)
 		}),
 	}
 )
@@ -83,7 +83,7 @@ func WithBackoffContext(bf BackoffFuncContext) CallOption {
 
 // WithOnRetryCallback sets the callback to use when a retry occurs.
 //
-// By default, no action is performed on retry.
+// By default, when no callback function provided, we will just print a log to trace
 func WithOnRetryCallback(fn OnRetryCallback) CallOption {
 	return CallOption{applyFunc: func(o *options) {
 		o.onRetryCallback = fn
--- a/interceptors/retry/options.go
+++ b/interceptors/retry/options.go
@@ -26,7 +26,7 @@ var (
 		backoffFunc: BackoffFuncContext(func(ctx context.Context, attempt uint) time.Duration {
 			return BackoffLinearWithJitter(50*time.Millisecond /*jitter*/, 0.10)(attempt)
 		}),
-		onRetryCallback: OnRetryCallbackContext(func(ctx context.Context, attempt uint, err error) {
+		onRetryCallback: OnRetryCallback(func(ctx context.Context, attempt uint, err error) {
 			// By default we don't have any callback logic to execute
 		}),
 	}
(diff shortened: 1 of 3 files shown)


interceptors/retry/options.go
```diff
@@ -75,6 +87,22 @@ func WithBackoffContext(bf BackoffFuncContext) CallOption {
...
+		})
+	}}
+}
+
+// WithOnRetryCallbackContext sets the `OnRetryCallbackContext` used to add additional logic when retry occurs.
+func WithOnRetryCallbackContext(fn OnRetryCallbackContext) CallOption {
+	return CallOption{applyFunc: func(o *options) {
+		o.onRetryCallback = fn
+	}}
+}
```
johanbrandhorst, line 104: Re: merge these into one

interceptors/retry/retry.go
```diff
@@ -48,6 +48,7 @@ func UnaryClientInterceptor(optFuncs ...CallOption) grpc.UnaryClientInterceptor
 			if lastErr == nil {
 				return nil
 			}
+			callOpts.onRetryCallback(callCtx, attempt, lastErr)
 			logTrace(parentCtx, "grpc_retry attempt: %d, got err: %v", attempt, lastErr)
```
johanbrandhorst, line 52: Maybe we can make this log here the default retry callback behaviour while we're here?

interceptors/retry/options.go
```diff
@@ -45,6 +48,15 @@ type BackoffFunc func(attempt uint) time.Duration
 // with the next iteration. The context can be used to extract request scoped metadata and context values.
 type BackoffFuncContext func(ctx context.Context, attempt uint) time.Duration
 
+// OnRetryCallback called in retry attempts flow, which can be used to add additiional logic when retry occurs
+//
+type OnRetryCallback func(attempt uint, err error)
+
+// OnRetryCallback called in retry attempts flow, which can be used to add additiional logic when retry occurs
+//
+// The context can be used to extract request scoped metadata and context values.
```
johanbrandhorst, line 57: After merging these, lets simplify this comment a bit too.

```suggestion
// OnRetryCallback is the type of function called when a retry occurs.
```

interceptors/retry/options.go
```diff
@@ -75,6 +87,22 @@ func WithBackoffContext(bf BackoffFuncContext) CallOption {
 	}}
 }
 
+// WithOnRetryCallback sets the `OnRetryCallback` used to add additional logic when retry occurs.
```
johanbrandhorst, line 90:
```suggestion
// WithOnRetryCallback sets the callback to use when a retry occurs.
//
// By default, no action is performed on retry.
```

johanbrandhorst (changes requested):
Thanks for taking this on! A few suggestions.

interceptors/retry/retry.go
shamil: I can do it, but If we do this, we will lose this trace log once the custom callback will be set bu user.

shamil:
Thanks for review @johanbrandhorst
Addressed all the comments, had only one concern about making the trace log the default callback behaviour. Let me know how you want to proceed and I will make the require changes.

interceptors/retry/retry.go
johanbrandhorst: We'll document that as the default behaviour so that the user knows that it will disappear if they override the callback. They can always manually add it back again.

shamil: I can also put this in the `WithOnRetryCallback` before invoking the user's logic, if this makes sense.

johanbrandhorst: No I think one of the things I want to do here is allow the user to disable this log behaviour too.

shamil: All right. Done

johanbrandhorst (approved):
LGTM

johanbrandhorst:
Thanks for your contribution!

Issue #412, opened 2021-04-21
v2: Tests are broked on v2 branch


ash2k:
https://github.com/grpc-ecosystem/go-grpc-middleware/runs/2203038370?check_suite_focus=true#step:5:83

bwplotka:
Thanks, cc @yashrsharma44 

We plan to do last cleanup and pass before moving with full v2, so all feedback is welcome for v2!

Issue #411, opened 2021-04-21
v2: Not all tests are executed on macOS


ash2k:
https://github.com/grpc-ecosystem/go-grpc-middleware/runs/2203038390?check_suite_focus=true#step:5:6

There is an issue in the makefile.

yashrsharma44:
This issue has been addressed by #413 so closing.
Let me know if this needs to be opened :)

ash2k:
I'm sorry, but it was not addressed - I was not fixing it. It's still happening e.g. https://github.com/grpc-ecosystem/go-grpc-middleware/runs/1895406088?check_suite_focus=true#step:5:7

ash2k:
Ok, here is a recent test run https://github.com/grpc-ecosystem/go-grpc-middleware/runs/2436081775?check_suite_focus=true#step:5:6

bwplotka:
We can't see logs (too old) ): Let's make sure to paste exact problem next time.

aboryslawski:
All tests pass for both `make test` and `go test ./...` for me locally. Apple M1 Pro macOS 13.6 (22G120).

aboryslawski:
Unless @ash2k is referring to some specific tests that are omitted.

ash2k:
Well, I don't remember what the problem was.

PR #413, opened 2021-04-21
v2: Refactor metrics interceptor and fix tests

ash2k:
I decided to have a look at the v2 branch as I'm interested in better Prometheus metrics. I found that stuff was in a very weird state so I tried to fix, simplify and clean it all up. See inline comments for some details. I have also found that tests fail and are not executed on macOS (https://github.com/grpc-ecosystem/go-grpc-middleware/issues/411).

Closes https://github.com/grpc-ecosystem/go-grpc-middleware/issues/412.

### DIFF
--- a/interceptors/client.go
+++ b/interceptors/client.go
@@ -7,6 +7,7 @@ package interceptors
 
 import (
 	"context"
+	"io"
 	"time"
 
 	"google.golang.org/grpc"
@@ -74,7 +75,10 @@ func (s *monitoredClientStream) RecvMsg(m interface{}) error {
 	if err == nil {
 		return nil
 	}
-
-	s.reporter.PostCall(err, time.Since(s.startTime))
+	var postErr error
+	if err != io.EOF {
+		postErr = err
+	}
+	s.reporter.PostCall(postErr, time.Since(s.startTime))
 	return err
 }
--- a/interceptors/client_test.go
+++ b/interceptors/client_test.go
@@ -50,7 +50,7 @@ func (m *mockReportable) Equal(t *testing.T, expected []*mockReport) {
 				require.NoError(t, err)
 				continue
 			}
-			require.Equal(t, expected[i].postCalls[k].Error(), err.Error(), "%v %v", i, k)
+			require.EqualError(t, err, expected[i].postCalls[k].Error(), "%v %v", i, k)
 		}
 		require.Len(t, expected[i].postMsgSends, len(e.postMsgSends), "%v", i)
 		for k, err := range e.postMsgSends {
@@ -276,7 +276,7 @@ func (s *ClientInterceptorTestSuite) TestListReporting() {
 		typ:             ServerStream,
 		svcName:         testpb.TestServiceFullName,
 		methodName:      "PingList",
-		postCalls:       []error{io.EOF},
+		postCalls:       []error{nil},
 		postMsgReceives: append(make([]error, testpb.ListResponseCount), io.EOF),
 		postMsgSends:    []error{nil},
 	}})
@@ -329,7 +329,9 @@ func (s *ClientInterceptorTestSuite) TestBiStreamingReporting() {
 			if err == io.EOF {
 				break
 			}
-			require.NoError(s.T(), err, "reading pingStream shouldn't fail")
+			if !s.Assert().NoError(err, "reading pingStream shouldn't fail") {
+				break
+			}
 			count++
 		}
 	}()
@@ -340,12 +342,12 @@ func (s *ClientInterceptorTestSuite) TestBiStreamingReporting() {
 	require.NoError(s.T(), ss.CloseSend())
 	wg.Wait()
 
-	require.EqualValues(s.T(), count, 100, "Number of received msg on the wire must match")
+	require.EqualValues(s.T(), 100, count, "Number of received msg on the wire must match")
 	s.mock.Equal(s.T(), []*mockReport{{
 		typ:             BidiStream,
 		svcName:         testpb.TestServiceFullName,
 		methodName:      "PingStream",
-		postCalls:       []error{io.EOF},
+		postCalls:       []error{nil},
 		postMsgReceives: append(make([]error, 100), io.EOF),
 		postMsgSends:    make([]error, 100),
 	}})
--- a/providers/openmetrics/client_interceptor.go
+++ b/providers/openmetrics/client_interceptor.go
@@ -1,35 +1,21 @@
 package metrics
 
 import (
-	openmetrics "github.com/prometheus/client_golang/prometheus"
 	"google.golang.org/grpc"
 
 	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors"
 )
 
-// RegisterClientMetrics returns a custom ClientMetrics object registered
-// with the user's registry, and registers some common metrics associated
-// with every instance.
-func RegisterClientMetrics(registry openmetrics.Registerer) *ClientMetrics {
-	customClientMetrics := NewClientMetrics(registry)
-	customClientMetrics.MustRegister(customClientMetrics.clientStartedCounter)
-	customClientMetrics.MustRegister(customClientMetrics.clientHandledCounter)
-	customClientMetrics.MustRegister(customClientMetrics.clientStreamMsgReceived)
-	customClientMetrics.MustRegister(customClientMetrics.clientStreamMsgSent)
-
-	return customClientMetrics
-}
-
 // UnaryClientInterceptor is a gRPC client-side interceptor that provides Prometheus monitoring for Unary RPCs.
-func UnaryClientInterceptor(clientRegister openmetrics.Registerer) grpc.UnaryClientInterceptor {
+func UnaryClientInterceptor(clientMetrics *ClientMetrics) grpc.UnaryClientInterceptor {
 	return interceptors.UnaryClientInterceptor(&reportable{
-		registry: clientRegister,
+		clientMetrics: clientMetrics,
 	})
 }
 
 // StreamClientInterceptor is a gRPC client-side interceptor that provides Prometheus monitoring for Streaming RPCs.
-func StreamClientInterceptor(clientRegister openmetrics.Registerer) grpc.StreamClientInterceptor {
+func StreamClientInterceptor(clientMetrics *ClientMetrics) grpc.StreamClientInterceptor {
 	return interceptors.StreamClientInterceptor(&reportable{
-		registry: clientRegister,
+		clientMetrics: clientMetrics,
 	})
 }
--- a/providers/openmetrics/client_metrics.go
+++ b/providers/openmetrics/client_metrics.go
@@ -1,99 +1,103 @@
 package metrics
 
 import (
-	"github.com/prometheus/client_golang/prometheus"
 	openmetrics "github.com/prometheus/client_golang/prometheus"
 )
 
 // ClientMetrics represents a collection of metrics to be registered on a
 // Prometheus metrics registry for a gRPC client.
 type ClientMetrics struct {
-	clientRegister openmetrics.Registerer
-
 	clientStartedCounter    *openmetrics.CounterVec
 	clientHandledCounter    *openmetrics.CounterVec
 	clientStreamMsgReceived *openmetrics.CounterVec
 	clientStreamMsgSent     *openmetrics.CounterVec
 
-	clientHandledHistogramEnabled bool
-	clientHandledHistogramOpts    openmetrics.HistogramOpts
-	clientHandledHistogram        *openmetrics.HistogramVec
-
-	clientStreamRecvHistogramEnabled bool
-	clientStreamRecvHistogramOpts    openmetrics.HistogramOpts
-	clientStreamRecvHistogram        *openmetrics.HistogramVec
-
-	clientStreamSendHistogramEnabled bool
-	clientStreamSendHistogramOpts    openmetrics.HistogramOpts
-	clientStreamSendHistogram        *openmetrics.HistogramVec
+	// clientHandledHistogram can be nil
+	clientHandledHistogram *openmetrics.HistogramVec
+	// clientStreamRecvHistogram can be nil
+	clientStreamRecvHistogram *openmetrics.HistogramVec
+	// clientStreamSendHistogram can be nil
+	clientStreamSendHistogram *openmetrics.HistogramVec
 }
 
-// NewClientMetrics returns a ClientMetrics object. Use a new instance of
-// ClientMetrics when not using the default Prometheus metrics registry, for
-// example when wanting to control which metrics are added to a registry as
-// opposed to automatically adding metrics via init functions.
-func NewClientMetrics(clientRegistry prometheus.Registerer, counterOpts ...CounterOption) *ClientMetrics {
-	opts := counterOptions(counterOpts)
+// NewClientMetrics returns a new ClientMetrics object.
+func NewClientMetrics(opts ...ClientMetricsOption) *ClientMetrics {
+	var config clientMetricsConfig
+	config.apply(opts)
 	return &ClientMetrics{
-		clientRegister: clientRegistry,
 		clientStartedCounter: openmetrics.NewCounterVec(
-			opts.apply(openmetrics.CounterOpts{
+			config.counterOpts.apply(openmetrics.CounterOpts{
 				Name: "grpc_client_started_total",
 				Help: "Total number of RPCs started on the client.",
 			}), []string{"grpc_type", "grpc_service", "grpc_method"}),
 
 		clientHandledCounter: openmetrics.NewCounterVec(
-			opts.apply(openmetrics.CounterOpts{
+			config.counterOpts.apply(openmetrics.CounterOpts{
 				Name: "grpc_client_handled_total",
 				Help: "Total number of RPCs completed by the client, regardless of success or failure.",
 			}), []string{"grpc_type", "grpc_service", "grpc_method", "grpc_code"}),
 
 		clientStreamMsgReceived: openmetrics.NewCounterVec(
-			opts.apply(openmetrics.CounterOpts{
+			config.counterOpts.apply(openmetrics.CounterOpts{
 				Name: "grpc_client_msg_received_total",
 				Help: "Total number of RPC stream messages received by the client.",
 			}), []string{"grpc_type", "grpc_service", "grpc_method"}),
 
 		clientStreamMsgSent: openmetrics.NewCounterVec(
-			opts.apply(openmetrics.CounterOpts{
+			config.counterOpts.apply(openmetrics.CounterOpts{
 				Name: "grpc_client_msg_sent_total",
 				Help: "Total number of gRPC stream messages sent by the client.",
 			}), []string{"grpc_type", "grpc_service", "grpc_method"}),
 
-		clientHandledHistogramEnabled: false,
-		clientHandledHistogramOpts: openmetrics.HistogramOpts{
-			Name:    "grpc_client_handling_seconds",
-			Help:    "Histogram of response latency (seconds) of the gRPC until it is finished by the application.",
-			Buckets: openmetrics.DefBuckets,
-		},
-		clientHandledHistogram:           nil,
-		clientStreamRecvHistogramEnabled: false,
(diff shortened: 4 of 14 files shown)


ash2k:
@googlebot I signed it!

interceptors/client.go
```diff
@@ -74,7 +75,10 @@ func (s *monitoredClientStream) RecvMsg(m interface{}) error {
 	if err == nil {
 		return nil
 	}
-
-	s.reporter.PostCall(err, time.Since(s.startTime))
+	var postErr error
+	if err != io.EOF {
+		postErr = err
+	}
+	s.reporter.PostCall(postErr, time.Since(s.startTime))
```
ash2k, line 82: The problem with passing `io.EOF` is that it's not a valid gRPC status code. The code there then turns it into status `Unknown` rather than `OK`, which is incorrect. A test was failing because of that.

interceptors/client_test.go
```diff
@@ -50,7 +50,7 @@ func (m *mockReportable) Equal(t *testing.T, expected []*mockReport) {
 				require.NoError(t, err)
 				continue
 			}
-			require.Equal(t, expected[i].postCalls[k].Error(), err.Error(), "%v %v", i, k)
+			require.EqualError(t, err, expected[i].postCalls[k].Error(), "%v %v", i, k)
```
ash2k, line 53: If there is a bug and `err` is `nil`, then this line panics. Now it does not and fails the test cleanly.

interceptors/client_test.go
```diff
@@ -276,7 +276,7 @@ func (s *ClientInterceptorTestSuite) TestListReporting() {
 		typ:             ServerStream,
 		svcName:         testpb.TestServiceFullName,
 		methodName:      "PingList",
-		postCalls:       []error{io.EOF},
+		postCalls:       []error{nil},
```
ash2k, line 279: Because of the change in `interceptors/client.go`, it's called with `nil` now.

interceptors/client_test.go
```diff
@@ -329,7 +329,9 @@ func (s *ClientInterceptorTestSuite) TestBiStreamingReporting() {
 			if err == io.EOF {
 				break
 			}
-			require.NoError(s.T(), err, "reading pingStream shouldn't fail")
```
ash2k, line 332: It is incorrect to use `require` in a goroutine other than the main test goroutine.

interceptors/client_test.go
```diff
@@ -340,12 +342,12 @@ func (s *ClientInterceptorTestSuite) TestBiStreamingReporting() {
 	require.NoError(s.T(), ss.CloseSend())
 	wg.Wait()
 
-	require.EqualValues(s.T(), count, 100, "Number of received msg on the wire must match")
+	require.EqualValues(s.T(), 100, count, "Number of received msg on the wire must match")
```
ash2k, line 345: Put actual and expected values where they should be.

providers/openmetrics/client_interceptor.go
```diff
@@ -1,35 +1,21 @@
...
-	openmetrics "github.com/prometheus/client_golang/prometheus"
 	"google.golang.org/grpc"
 
 	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors"
 )
 
-// RegisterClientMetrics returns a custom ClientMetrics object registered
-// with the user's registry, and registers some common metrics associated
-// with every instance.
-func RegisterClientMetrics(registry openmetrics.Registerer) *ClientMetrics {
```
ash2k, line 13: Moved and simplified.

providers/openmetrics/client_interceptor.go
```diff
@@ -1,35 +1,21 @@
...
-	customClientMetrics.MustRegister(customClientMetrics.clientHandledCounter)
-	customClientMetrics.MustRegister(customClientMetrics.clientStreamMsgReceived)
-	customClientMetrics.MustRegister(customClientMetrics.clientStreamMsgSent)
-
-	return customClientMetrics
-}
-
 // UnaryClientInterceptor is a gRPC client-side interceptor that provides Prometheus monitoring for Unary RPCs.
-func UnaryClientInterceptor(clientRegister openmetrics.Registerer) grpc.UnaryClientInterceptor {
+func UnaryClientInterceptor(clientMetrics *ClientMetrics) grpc.UnaryClientInterceptor {
```
ash2k, line 10: This was really weird. Interceptor shouldn't care about the registry, it only needs to increment the metrics. It shouldn't need to create it's own metrics and register them either.

providers/openmetrics/client_metrics.go
```diff
@@ -1,99 +1,103 @@
...
-
-	clientStreamSendHistogramEnabled bool
-	clientStreamSendHistogramOpts    openmetrics.HistogramOpts
-	clientStreamSendHistogram        *openmetrics.HistogramVec
+	// clientHandledHistogram can be nil
+	clientHandledHistogram *openmetrics.HistogramVec
+	// clientStreamRecvHistogram can be nil
+	clientStreamRecvHistogram *openmetrics.HistogramVec
+	// clientStreamSendHistogram can be nil
+	clientStreamSendHistogram *openmetrics.HistogramVec
```
ash2k, line 20: Use `nil` a the flag rather than a separate field. Better memory alignment and smaller object size.

providers/openmetrics/client_metrics.go
```diff
@@ -1,99 +1,103 @@
...
 
 import (
-	"github.com/prometheus/client_golang/prometheus"
 	openmetrics "github.com/prometheus/client_golang/prometheus"
 )
 
 // ClientMetrics represents a collection of metrics to be registered on a
 // Prometheus metrics registry for a gRPC client.
 type ClientMetrics struct {
-	clientRegister openmetrics.Registerer
```
ash2k, line 11: We don't need it (anymore).

ash2k, line 11: We don't need it (anymore).

providers/openmetrics/client_metrics.go
```diff
@@ -1,99 +1,103 @@
...
 }
 
-// NewClientMetrics returns a ClientMetrics object. Use a new instance of
-// ClientMetrics when not using the default Prometheus metrics registry, for
-// example when wanting to control which metrics are added to a registry as
-// opposed to automatically adding metrics via init functions.
-func NewClientMetrics(clientRegistry prometheus.Registerer, counterOpts ...CounterOption) *ClientMetrics {
-	opts := counterOptions(counterOpts)
+// NewClientMetrics returns a new ClientMetrics object.
+func NewClientMetrics(opts ...ClientMetricsOption) *ClientMetrics {
```
ash2k, line 24: Constructor takes options for customization. Options can enable histograms and provide other configuration.

providers/openmetrics/client_metrics.go
```diff
@@ -123,65 +127,13 @@ func (m *ClientMetrics) Collect(ch chan<- openmetrics.Metric) {
...
-	if m.clientStreamRecvHistogramEnabled {
+	if m.clientStreamRecvHistogram != nil {
 		m.clientStreamRecvHistogram.Collect(ch)
 	}
-	if m.clientStreamSendHistogramEnabled {
+	if m.clientStreamSendHistogram != nil {
 		m.clientStreamSendHistogram.Collect(ch)
 	}
 }
-
```
ash2k, line 136: All of that has been turned into options for the constructor. `ClientMetrics` is now immutable (from the outside) and there is no chance of a data race because someone calls these methods.

providers/openmetrics/client_test.go
```diff
@@ -7,17 +7,13 @@ import (
...
-	"github.com/prometheus/client_golang/prometheus"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
-)
 
-var (
-	DefaultClientMetrics *ClientMetrics = NewClientMetrics(prometheus.DefaultRegisterer)
```
ash2k, line 20: Moved to the suite, no need for package-global.

providers/openmetrics/client_test.go
```diff
@@ -33,12 +29,13 @@ type ClientInterceptorTestSuite struct {
...
 	ctx            context.Context
 	cancel         context.CancelFunc
+	clientMetrics  *ClientMetrics
 }
 
 func (s *ClientInterceptorTestSuite) SetupSuite() {
 	var err error
 
-	DefaultClientMetrics.EnableClientHandlingTimeHistogram()
+	s.clientMetrics = NewClientMetrics(WithClientHandlingTimeHistogram())
```
ash2k, line 38: We don't even need to register this with a registry now - we test how interceptors increment counters/etc, don't care about the registry.

providers/openmetrics/client_test.go
```diff
@@ -67,22 +64,21 @@ func (s *ClientInterceptorTestSuite) SetupTest() {
...
+	s.clientMetrics.clientHandledCounter.Reset()
+	s.clientMetrics.clientHandledHistogram.Reset()
+	s.clientMetrics.clientStreamMsgReceived.Reset()
+	s.clientMetrics.clientStreamMsgSent.Reset()
 }
 
 func (s *ClientInterceptorTestSuite) TearDownSuite() {
+	if s.clientConn != nil {
+		s.clientConn.Close()
+	}
```
ash2k, line 77: Close the client first, then the server for a clean shutdown.

providers/openmetrics/client_test.go
```diff
@@ -93,33 +89,34 @@ func (s *ClientInterceptorTestSuite) TearDownTest() {
 func (s *ClientInterceptorTestSuite) TestUnaryIncrementsMetrics() {
 	_, err := s.testClient.PingEmpty(s.ctx, &pb_testproto.PingEmptyRequest{}) // should return with code=OK
 	require.NoError(s.T(), err)
-	requireValue(s.T(), 1, DefaultClientMetrics.clientStartedCounter.WithLabelValues("unary", "mwitkow.testproto.TestService", "PingEmpty"))
-	requireValue(s.T(), 1, DefaultClientMetrics.clientHandledCounter.WithLabelValues("unary", "mwitkow.testproto.TestService", "PingEmpty", "OK"))
-	requireValueHistCount(s.T(), 1, DefaultClientMetrics.clientHandledHistogram.WithLabelValues("unary", "mwitkow.testproto.TestService", "PingEmpty"))
+	requireValue(s.T(), 1, s.clientMetrics.clientStartedCounter.WithLabelValues("unary", "providers.openmetrics.testproto.v1.TestService", "PingEmpty"))
+	requireValue(s.T(), 1, s.clientMetrics.clientHandledCounter.WithLabelValues("unary", "providers.openmetrics.testproto.v1.TestService", "PingEmpty", "OK"))
+	requireValueHistCount(s.T(), 1, s.clientMetrics.clientHandledHistogram.WithLabelValues("unary", "providers.openmetrics.testproto.v1.TestService", "PingEmpty"))
```
ash2k, line 94: Package name was wrong. How did that get merged? :)

providers/openmetrics/client_test.go
```diff
@@ -93,33 +89,34 @@ func (s *ClientInterceptorTestSuite) TearDownTest() {
...
 	_, err = s.testClient.PingList(s.ctx, &pb_testproto.PingListRequest{ErrorCodeReturned: uint32(codes.FailedPrecondition)}) // should return with code=FailedPrecondition
 	require.NoError(s.T(), err, "PingList must not fail immediately")
-	requireValue(s.T(), 2, DefaultClientMetrics.clientStartedCounter.WithLabelValues("server_stream", "mwitkow.testproto.TestService", "PingList"))
+	requireValue(s.T(), 2, s.clientMetrics.clientStartedCounter.WithLabelValues("server_stream", "providers.openmetrics.testproto.v1.TestService", "PingList"))
 }
 
 func (s *ClientInterceptorTestSuite) TestStreamingIncrementsMetrics() {
-	ss, _ := s.testClient.PingList(s.ctx, &pb_testproto.PingListRequest{}) // should return with code=OK
+	ss, err := s.testClient.PingList(s.ctx, &pb_testproto.PingListRequest{}) // should return with code=OK
+	require.NoError(s.T(), err)
```
ash2k, line 115: Missing check

providers/openmetrics/client_test.go
```diff
@@ -128,21 +125,21 @@ func (s *ClientInterceptorTestSuite) TestStreamingIncrementsMetrics() {
...
-	requireValue(s.T(), 1, DefaultClientMetrics.clientHandledCounter.WithLabelValues("server_stream", "mwitkow.testproto.TestService", "PingList", "OK"))
-	requireValue(s.T(), countListResponses, DefaultClientMetrics.clientStreamMsgReceived.WithLabelValues("server_stream", "mwitkow.testproto.TestService", "PingList"))
-	requireValue(s.T(), 1, DefaultClientMetrics.clientStreamMsgSent.WithLabelValues("server_stream", "mwitkow.testproto.TestService", "PingList"))
-	requireValueHistCount(s.T(), 1, DefaultClientMetrics.clientHandledHistogram.WithLabelValues("server_stream", "mwitkow.testproto.TestService", "PingList"))
+	requireValue(s.T(), 1, s.clientMetrics.clientStartedCounter.WithLabelValues("server_stream", "providers.openmetrics.testproto.v1.TestService", "PingList"))
+	requireValue(s.T(), 1, s.clientMetrics.clientHandledCounter.WithLabelValues("server_stream", "providers.openmetrics.testproto.v1.TestService", "PingList", "OK"))
+	requireValue(s.T(), countListResponses+1 /* + EOF */, s.clientMetrics.clientStreamMsgReceived.WithLabelValues("server_stream", "providers.openmetrics.testproto.v1.TestService", "PingList"))
```
ash2k, line 130: It takes 1 extra call to get `io.EOF` from the stream, and that call is also counted. So it's +1 here. How did that get merged?

providers/openmetrics/go.mod
```diff
@@ -11,3 +11,5 @@ require (
 	google.golang.org/grpc v1.35.0
 	google.golang.org/protobuf v1.25.0
 )
+
+replace github.com/grpc-ecosystem/go-grpc-middleware/v2 => ../..
```
ash2k, line 15: I'm not sure how to test it otherwise if changes are made to the outer module and to providers.

providers/openmetrics/reporter.go
```diff
@@ -61,17 +61,16 @@ func (r *reporter) PostMsgReceive(_ interface{}, _ error, _ time.Duration) {
 }
 
 type reportable struct {
-	registry openmetrics.Registerer
+	clientMetrics *ClientMetrics
+	serverMetrics *ServerMetrics
 }
 
 func (rep *reportable) ServerReporter(ctx context.Context, _ interface{}, typ interceptors.GRPCType, service string, method string) (interceptors.Reporter, context.Context) {
-	m := NewServerMetrics(rep.registry)
```
ash2k, line 68: This is super weird. So we're creating a new metrics holder here, we do not register it anywhere and just use it as is. This is just broken - there is no way these metrics are going to be scraped.

providers/openmetrics/server_interceptor.go
```diff
@@ -1,35 +1,21 @@
 package metrics
```
ash2k, line 1: Server metrics and interceptor are refactored in the same way as client metrics and interceptor. All comments apply, not repeating them.

providers/openmetrics/server_test.go
```diff
@@ -167,36 +168,36 @@ func (s *ServerInterceptorTestSuite) TestStreamingIncrementsMetrics() {
...
 // would have while scraping this endpoint.
 // Order of matching label vales does not matter.
-func fetchPrometheusLines(t *testing.T, metricName string, matchingLabelValues ...string) []string {
+func fetchPrometheusLines(t *testing.T, reg prometheus.Gatherer, metricName string, matchingLabelValues ...string) []string {
 	resp := httptest.NewRecorder()
 	req, err := http.NewRequest("GET", "/", nil)
 	require.NoError(t, err, "failed creating request for Prometheus handler")
 
-	promhttp.Handler().ServeHTTP(resp, req)
+	promhttp.HandlerFor(reg, promhttp.HandlerOpts{}).ServeHTTP(resp, req)
```
ash2k, line 200: Use the explicitly passed registry here rather than a global one. Test should be self-contained and hermetic, it shouldn't depend on globals.

providers/openmetrics/server_test.go
```diff
@@ -292,6 +293,7 @@ func toFloat64HistCount(h prometheus.Observer) uint64 {
 }
 
 func requireValue(t *testing.T, expect int, c prometheus.Collector) {
+	t.Helper()
```
ash2k, line 296: That way error traces when a test fails lead to the line where this function was invoked, not into this function. Convenience thing, helped me to navigate failing tests.

interceptors/client.go
johanbrandhorst: `io.EOF` is returned from `RecvMsg` to indicate the end of the stream (https://pkg.go.dev/google.golang.org/grpc#ClientStream), does this no break that behaviour?

providers/openmetrics/client_test.go
johanbrandhorst: Nice find

providers/openmetrics/go.mod
johanbrandhorst: I think it'll have to be merged in two steps.

johanbrandhorst (commented):
This looks generally good to me, thanks for taking the time. I'll hold off on approving as this is @bwplotka's area of expertise.

interceptors/client.go
ash2k: No it does not - note that the function still returns `err` i.e. `io.EOF`, we only change what we pass to `PostCall()`.

providers/openmetrics/go.mod
ash2k: If this PR is approved I can split it into two - one updating the base and another one updating the provider. I also have a few more things I'd like to cleanup in follow ups.

interceptors/client.go
johanbrandhorst: Fair enough, makes sense.

providers/openmetrics/go.mod
johanbrandhorst: Sounds good to me, will give @bwplotka a few days to have a look but I'm otherwise happy to merge this on our own.

interceptors/client_test.go
bwplotka: 👍🏽 Thanks a lot

providers/openmetrics/client_interceptor.go
bwplotka: Agree, cc @yashrsharma44 for learning experience (:

interceptors/client_test.go
bwplotka: Wow, thanks for those detailed comments (: Next time, no need for this it's very clear! 💪🏽

providers/openmetrics/go.mod
bwplotka: Yup, this is the problem with multi-modules.

One thing we could add is to add some tooling that e.g makes commit with this PR on your fork, then creates commit on top of it, so it's at least not broken on master, if you just pull openmetrics module. Local path approach can break users on main. It's fine for now (dev part of v2) but we might want to improve this process. LGTM for now

providers/openmetrics/server_test.go
bwplotka: I agree, thanks for this

bwplotka (approved):
Thanks for this, LGTM

I agree that the openmetrics PR might have been [prematurely merged.](https://github.com/grpc-ecosystem/go-grpc-middleware/pull/379). Yet it was kind of stated to merge a bigger chunk of work for later to iterate... yet iteration never happened and the code on main was unusable. It's totally on me, sorry for not paying close attention during the review process. Cc @yashrsharma44 for learning too on this (: 

image

Let's make sure to keep the quality bar much higher next time.

yashrsharma44:
Thanks, @ash2k for such great work!

I was the original author for introducing the changes for openmetrics and I agree that the changes were not suitable to be merged. Thanks for such a comprehensive review and this PR might be a good opportunity for me to write good Go code 😄

PR #421, opened 2021-05-13
Simplify chain interceptors

aarongable:
The four (client/server, unary/stream) interceptors have to wrap a slice
of interceptors in functions which satisfy the handler interface, but
which are closures over the other parameters an interceptor is expected
to have.

The previous approach accomplished this goal with recursion. This had
two drawbacks: first, the code was difficult to understand, as most
recursion attempts to fully encode state in function parameters, but
here state was necessarily also encoded in the closure; and second,
the recursive base-case meant that even the innermost interceptor was
not calling the bare handler, it was calling a wrapped handler.

This new approach instead iteratively constructs wrappers from the
inside out. It results in fewer lines of code, with fewer variables
held in each closure. Hopefully this leads to higher readability.
It also makes it so that the innermost interceptor is calling the gRPC
handler directly, allowing introspection using `reflect.ValueOf(handler)`
to return the actual underlying handler instead of an anonymous func.

### DIFF
--- a/chain.go
+++ b/chain.go
@@ -19,19 +19,36 @@ import (
 func ChainUnaryServer(interceptors ...grpc.UnaryServerInterceptor) grpc.UnaryServerInterceptor {
 	n := len(interceptors)
 
+	// Dummy interceptor maintained for backward compatibility to avoid returning nil.
+	if n == 0 {
+		return func(ctx context.Context, req interface{}, _ *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+			return handler(ctx, req)
+		}
+	}
+
+	// The degenerate case, just return the single wrapped interceptor directly.
+	if n == 1 {
+		return interceptors[0]
+	}
+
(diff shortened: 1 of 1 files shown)


aarongable:
@googlebot I signed it!

johanbrandhorst (approved):
Hi Aaron, thanks for your contribution. I want to note that this logic is going to be abandoned soon for two reasons:

1. We're working on v2 of the middleware, which does not use this code because...
2. There is an official implementation of this in the grpc-go library. See https://pkg.go.dev/google.golang.org/grpc#WithChainStreamInterceptor etc.

With that in mind, are you still interested in contributing this? The change looks good to me, and thanks for adding all the colorful comments :).

aarongable:
Apologies for the delay! Yes, I'm still happy to contribute this; as far as I'm concerned its much more up to you as to whether you want churn in this code shortly before it is removed :) Happy for this to be merged, or for it to be closed, at your discretion!

johanbrandhorst:
I'm happy enough to merge it, thanks for your contribution!

PR #425, opened 2021-06-14
v2: Add support for the phuslog logging provider

ogimenezb:
Integrates phuslog into go-grpc-middleware/providers.

I welcome all feedback!


ogimenezb:
@googlebot I fixed it.

ogimenezb:
@googlebot I fixed it.

providers/phuslog/logger.go
```diff
@@ -0,0 +1,49 @@
...
+	case logging.INFO:
+		l.Info(msg)
+	case logging.WARNING:
+		l.Warning(msg)
+	case logging.ERROR:
+		l.Error(msg)
+	default:
+		// TODO(kb): Perhaps this should be a logged warning, defaulting to ERROR to get attention
+		// without interrupting code flow?
+		panic(fmt.Sprintf("phuslog: unknown level %s", lvl))
```
johanbrandhorst, line 38: Yeah I don't think this sounds reasonable, indeed maybe just fall back to error?

johanbrandhorst (changes requested):
Except for the logging behavior when using an unknown log level, this LGTM.

providers/phuslog/logger.go
ogimenezb:
```suggestion
		l.NoLevelErr(fmt.Errorf("phuslog: unknown level %s", lvl), msg)
```

I have just implemented a `NoLevelErr` which would output
```go
{"level":"error","error":"phuslog: unknown level AAA","message":"a message log"}
```

Given that we are logging with unknown level I thinks it's best to choose Error Level because is the minimum any user would have...

If you agree I will push changes algo to phuslog

johanbrandhorst: Only change this one, and yes fall back to error.

johanbrandhorst (approved):
LGTM

PR #429, opened 2021-06-18
logging/zap/ctxzap: add caller skip to shorthand functions

jkawamoto:
The caller annotations made by the shorthand logging functions are`ctxzap/context.go:xx` regardless of where the shorthands are called. This commit adds caller skip so that each log message has the correct caller annotation.


johanbrandhorst (approved):
LGTM

PR #431, opened 2021-07-01
grpc_zap: fix caller annotation

jkawamoto:
Since `zapGrpcLoggerV2` wraps a zap's logger, caller annotations always point `grpclogger.go` instead of the expected file. This commit adds caller skip so that each log message has the correct annotation.

This commit also upgrades zap's version to v1.18.1 to use `zap.OnFatal` in tests. That function is necessary to catch fatal-level log messages.


johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

PR #432, opened 2021-07-01
v2:providers/zap: fix caller annotation

jkawamoto:
`Logger.Log` calls a zap's log function inside of this function. As a result, caller annotations always point `zap/logger.go`. This commit adds caller skip so that each log message has the correct annotation.


johanbrandhorst (approved):
LGTM

johanbrandhorst:
Thanks for your contribution!

PR #436, opened 2021-07-11
logging: Restore compatibility with v1 of pb Message

olivierlemasle:
Services generated with old versions of protoc-gen-go use the deprecated `github.com/golang/protobuf/proto.Message` (v1 messages) instead of `google.golang.org/protobuf/proto.Message` (v2 messages).

This commit makes the payload interceptor work with both versions.

### DIFF
--- a/interceptors/logging/payload.go
+++ b/interceptors/logging/payload.go
@@ -9,6 +9,9 @@ import (
 	"fmt"
 	"time"
 
+	//lint:ignore SA1019 we use this deprecated package to convert pb v1 messages to pb v2
+	//nolint:staticcheck // SA1019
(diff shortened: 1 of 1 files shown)


johanbrandhorst (commented):
I'm not sure this is something we actually want to support - this will only add support for pre v1.4.0 versions of golang/protobuf messages (right?), and a major goal of v2 is to move onto the Go APIv2 stack. If users are using < v1.4.0 of golang/protobuf, why don't they keep using v1?

bwplotka:
Yup, let's focus on middlewares v2 WDYT?

olivierlemasle:
Yes, I think you're right.

PR #438, opened 2021-07-13
Add all-validator support

leventeliu:
PGV changed its interface (again 😹) and a new interface `ValidateAll() error` was added to implement a partial function of the legacy interface `Validate(bool) error`.

I think it's a good idea to keep this middleware up-to-date and allow users to choose this new method `ValidateAll` over `Validate`. Also, backward compatibility is considered in these changes. Please review.


leventeliu:
@googlebot I signed it!

validator/validator.go
```diff
@@ -51,9 +78,13 @@ func UnaryServerInterceptor() grpc.UnaryServerInterceptor {
 // UnaryClientInterceptor returns a new unary client interceptor that validates outgoing messages.
 //
 // Invalid messages will be rejected with `InvalidArgument` before sending the request to server.
-func UnaryClientInterceptor() grpc.UnaryClientInterceptor {
+func UnaryClientInterceptor(validateAll ...bool) grpc.UnaryClientInterceptor {
```
johanbrandhorst, line 81: Why is this signature change needed? This is a potentially breaking change

validator/validator.go
```diff
@@ -11,6 +11,11 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+// The validateAller interface at protoc-gen-validate main branch.
```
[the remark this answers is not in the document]
leventeliu: Hi, here is the PR: [#468](https://github.com/envoyproxy/protoc-gen-validate/pull/468/files).

validator/validator.go
leventeliu: Hi, I don't see how this can break old code using zero arguments, and this is a common way to add (one) argument for a function without breaking things. Any better idea?

validator/validator.go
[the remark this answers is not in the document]
johanbrandhorst: I mean add it in a comment in the code, for future reference.

validator/validator.go
johanbrandhorst: Yeah this is a common assumption, but this is actually a breaking change if someone has assigned this function to a variable at some point:
```go
var interceptor func() grpc.UnaryClientInterceptor = validator.UnaryClientInterceptor
```
I think it's going to be very rarely cause an action breakage but it is technically breaking. Why do we need to add this parameter_

validator/validator.go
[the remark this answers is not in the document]
leventeliu: Done.

validator/validator.go
leventeliu: Ok, as I said in the PR description, I changed this signature to allow users to choose the new method `ValidateAll` over `Validate`.

As I know this new method uses a multi-error to return all validation failures. IMHO this would be much more friendly while designing user-oriented and non-performance sensitive APIs -- to tell the user which fields you provided are invalid, at one time.

And as you can see in the previous implementation, while using the legacy interface `Validate(all bool) error`, this middleware directly calls `Validate(false)`. In this way, it somehow hides part of PGV's functionality, which doesn't seem to be a good practice to me too.

So, here comes the point we should weigh the pros and cons of changing this signature. As you mentioned such kind of breaking is a real rare case, should we change it here? Or maybe we should create a new type of interceptor if I would like to use `ValidateAll`?

johanbrandhorst: I see now that this is actually targeting the master branch - I think we can just move this PR to the v2 branch and then we can change the signature without worrying about breaking users as it is still in development. What do you think?

leventeliu: Surely, I cherry-picked it into a new branch and opened a new PR #443. Now I will close this, thanks.

PR #439, opened 2021-07-13
recovery: change the default behavior

### LINKED ISSUE #435: v2: recovery default handler changes
Hi there.

I'd like to discus the [default](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/v2/interceptors/recovery/interceptors.go#L54) panic handler in recovery package.

From my experience we always have a top level error handling interceptor that catches all **non-grpc** errors, sends them to a service like sentry and hides the details from users with `status.Error(codes.Internal, "internal error")`, so at least it should be `fmt.Errorf("panic caught: %v", p)`.

The second thing is panic stack traces which are really important to track the root cause. The standard http server does it [this way](https://github.com/golang/go/blob/go1.16.5/src/net/http/server.go#L1820).

I do understand that this can be achieved using a custom panic handler but I suppose many people would like this changes and I would stop copy-pasting the following code to every project:

```go
func recoverPanic(p interface{}) error {
	stack := make([]byte, 64<<10)
	stack = stack[:runtime.Stack(stack, false)]
	return &PanicError{Panic: p, Stack: stack}
}

type PanicError struct {
	Panic interface{}
	Stack []byte
}

func (e *PanicError) Error() string {
	return fmt.Sprintf("panic caught: %v\n\n%s", e.Panic, e.Stack)
}
```

amenzhinsky:
Closes #435

### DIFF
--- a/interceptors/recovery/interceptors.go
+++ b/interceptors/recovery/interceptors.go
@@ -8,10 +8,10 @@ package recovery
 
 import (
 	"context"
+	"fmt"
+	"runtime"
 
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 )
 
 // RecoveryHandlerFunc is a function that recovers from the panic `p` by returning an `error`.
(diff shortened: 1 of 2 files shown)


johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

PR #442, opened 2021-07-13
v2:interceptors/logging: allow to separate request response payload logging

michaljemala:
Logging payloads of requests on the server-side and similarly of responses on the client-side is a pretty common and desired use-case. However, logging by default payloads of both requests as well as responses in both scenarios is IMO not a very common requirement.

Consider the following PR to enable more granular control of what to log and when. This is achieved by using a special `PayloadDecison` type having the following values:
* `NoPayloadLogging` - Payload logging is disabled.
* `LogRequest` - Only logging of requests is enabled.
* `LogResponse` - Only logging of responses is enabled.
* `LogRequestAndResponse` - Logging of both requests and responses is enabled.
The `PayloadDecison` is to be used instead of the original `bool` by the `logging.ServerPayloadLoggingDecider` and `logging.ClientPayloadLoggingDecider`.

NOTE: I quickly put this together to see whether it could be considered a valid approach, thus please excuse the lack of additional unit tests. I will be glad to add those once the approach is approved.

### DIFF
--- a/interceptors/logging/logging.go
+++ b/interceptors/logging/logging.go
@@ -67,15 +67,29 @@ func DefaultDeciderMethod(_ string, _ error) Decision {
 	return LogStartAndFinishCall
 }
 
+// PayloadDecision defines rules for enabling payload logging of request and responses.
+type PayloadDecision int
+
+const (
+	// NoPayloadLogging - Payload logging is disabled.
+	NoPayloadLogging PayloadDecision = iota
+	// LogPayloadRequest - Only logging of requests is enabled.
+	LogPayloadRequest
+	// LogPayloadResponse - Only logging of responses is enabled.
+	LogPayloadResponse
+	// LogPayloadRequestAndResponse - Logging of both requests and responses is enabled.
+	LogPayloadRequestAndResponse
+)
+
 // ServerPayloadLoggingDecider is a user-provided function for deciding whether to log the server-side
 // request/response payloads
-type ServerPayloadLoggingDecider func(ctx context.Context, fullMethodName string, servingObject interface{}) bool
+type ServerPayloadLoggingDecider func(ctx context.Context, fullMethodName string, servingObject interface{}) PayloadDecision
 
 // ClientPayloadLoggingDecider is a user-provided function for deciding whether to log the client-side
 // request/response payloads
-type ClientPayloadLoggingDecider func(ctx context.Context, fullMethodName string) bool
+type ClientPayloadLoggingDecider func(ctx context.Context, fullMethodName string) PayloadDecision
 
-// JsonPbMarshaller is a marshaller that serializes protobuf messages.
+// JsonPbMarshaler is a marshaller that serializes protobuf messages.
 type JsonPbMarshaler interface {
 	Marshal(pb proto.Message) ([]byte, error)
 }
--- a/interceptors/logging/payload.go
+++ b/interceptors/logging/payload.go
@@ -18,14 +18,20 @@ import (
 )
 
 type serverPayloadReporter struct {
-	ctx    context.Context
-	logger Logger
+	ctx      context.Context
+	logger   Logger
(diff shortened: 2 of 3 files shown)


michaljemala:
@googlebot I fixed it.

michaljemala:
@googlebot I fixed it.

johanbrandhorst (commented):
I quite like this, but it was @bwplotka who refactored these, so I would prefer to have his input.

interceptors/logging/payload.go
```diff
@@ -28,6 +29,9 @@ func (c *serverPayloadReporter) PostMsgSend(req interface{}, err error, duration
 	if err != nil {
 		return
 	}
+	if c.decision != LogResponse && c.decision != LogRequestAndResponse {
```
yashrsharma44, line 32: Can we use `switch case` in case we want to add more conditionals later on for different `PayloadDecision`? Would be easy to read and maintain the code 😄

yashrsharma44 (changes requested):
Functionally - I think using a switch case would make this code easier to read and maintainable. Otherwise looks good to me!

interceptors/logging/payload.go
michaljemala: Yes sure, good idea. I just changed it.

yashrsharma44:
Hi @michaljemala !

This PR looks good to me, can you run `make lint` for fixing the CI changes required?
Otherwise looks good to me.

yashrsharma44:
cc'ing @bwplotka for the inputs 😄

yashrsharma44:
Seems like the CI build needs approval for first time contributors - https://github.blog/changelog/2021-04-22-github-actions-maintainers-must-approve-first-time-contributor-workflow-runs/

michaljemala:
Hi @yashrsharma44 is there anything else on my side to be done?

yashrsharma44 (approved):

interceptors/logging/logging.go
```diff
@@ -67,15 +67,29 @@ func DefaultDeciderMethod(_ string, _ error) Decision {
...
 }
 
+// PayloadDecision defines rules for enabling payload logging of request and responses.
+type PayloadDecision int
+
+const (
+	// NoPayloadLogging - Payload logging is disabled.
+	NoPayloadLogging PayloadDecision = iota
+	// LogRequest - Only logging of requests is enabled.
+	LogRequest
```
bwplotka, line 77:
```suggestion
	LogPayloadRequest
```

interceptors/logging/logging.go
```diff
@@ -67,15 +67,29 @@ func DefaultDeciderMethod(_ string, _ error) Decision {
...
+	// LogResponse - Only logging of responses is enabled.
+	LogResponse
```
bwplotka, line 79:
```suggestion
	LogPayloadResponse
```

interceptors/logging/logging.go
```diff
@@ -67,15 +67,29 @@ func DefaultDeciderMethod(_ string, _ error) Decision {
...
+	// LogRequestAndResponse - Logging of both requests and responses is enabled.
+	LogRequestAndResponse
```
bwplotka, line 81:
```suggestion
	LogPayloadRequestAndResponse
```

bwplotka (approved):
LGTM, only minor nit and lint error 🤗 

Thanks!

interceptors/logging/logging.go
michaljemala: Done.

interceptors/logging/logging.go
michaljemala: Done.

interceptors/logging/logging.go
michaljemala: Done.

bwplotka (approved):
Thanks!

PR #443, opened 2021-07-16
Add all-validator support

leventeliu:
Reopen pull request #438 with new branch and target branch is `v2`.

FYI, I changed `PingErrorRequest` to implement only `validatorLegacy` interface, in order to cover more possible branches and keep coverage up. All tests passed and everything looks fine.

### DIFF
--- a/interceptors/validator/validator.go
+++ b/interceptors/validator/validator.go
@@ -11,6 +11,12 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+// The validateAller interface at protoc-gen-validate main branch.
+// See https://github.com/envoyproxy/protoc-gen-validate/pull/468.
+type validateAller interface {
+	ValidateAll() error
+}
+
 // The validate interface starting with protoc-gen-validate v0.6.0.
 // See https://github.com/envoyproxy/protoc-gen-validate/pull/455.
 type validator interface {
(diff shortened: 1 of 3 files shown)


interceptors/validator/validator.go
```diff
@@ -41,9 +63,13 @@ func validate(req interface{}) error {
 // UnaryServerInterceptor returns a new unary server interceptor that validates incoming messages.
 //
 // Invalid messages will be rejected with `InvalidArgument` before reaching any userspace handlers.
-func UnaryServerInterceptor() grpc.UnaryServerInterceptor {
+func UnaryServerInterceptor(validateAll ...bool) grpc.UnaryServerInterceptor {
```
johanbrandhorst, line 66: Do we need to make this variadic now? Seems we could just make it a single bool and force the user to supply it right? Could you please add what the meaning of the variable is to the comment above the method? Thanks!

interceptors/validator/validator.go
```diff
@@ -53,9 +79,13 @@ func UnaryServerInterceptor() grpc.UnaryServerInterceptor {
 // UnaryClientInterceptor returns a new unary client interceptor that validates outgoing messages.
 //
 // Invalid messages will be rejected with `InvalidArgument` before sending the request to server.
-func UnaryClientInterceptor() grpc.UnaryClientInterceptor {
+func UnaryClientInterceptor(validateAll ...bool) grpc.UnaryClientInterceptor {
```
johanbrandhorst, line 82: Same here

interceptors/validator/validator.go
```diff
@@ -68,22 +98,30 @@ func UnaryClientInterceptor() grpc.UnaryClientInterceptor {
 // type of the RPC. For `ServerStream` (1:m) requests, it will happen before reaching any userspace
 // handlers. For `ClientStream` (n:1) or `BidiStream` (n:m) RPCs, the messages will be rejected on
 // calls to `stream.Recv()`.
-func StreamServerInterceptor() grpc.StreamServerInterceptor {
+func StreamServerInterceptor(validateAll ...bool) grpc.StreamServerInterceptor {
```
johanbrandhorst, line 101: Same here

interceptors/validator/validator.go
leventeliu: Agree! Since we are working on v2, keeping the signature compatible with v1 is not actually necessary. Use a clearer single argument instead. Also add comments for the new argument.

Please review again. Thanks!

johanbrandhorst (approved):
LGTM!

PR #394, opened 2021-01-30
Removed tags; Simplified interceptor code; Added logging fields editability.

bwplotka:
Kind of biggy, but should be much cleaner.

The only surprise is that tracing is using `logging.Fields` to both fill span metadata (trace ID, sampled or not, span ID) as well. I think this is useful AND it was like that before 🤗 

At least it's a clean separation of responsibilities who is using what vs generic tags.

Fixes https://github.com/grpc-ecosystem/go-grpc-middleware/issues/382

NOTE:  I removed tracing package completely since in new PR we complete rewrite it anyway.

Signed-off-by: Bartlomiej Plotka <bwplotka@gmail.com>

PR #450, opened 2021-08-08
Refactor tracing interceptor

XSAM:
Split from #312


yashrsharma44:
Since #394 has been merged, there have been some changes in the structuring of `tags`(which are removed, replaced with `logging.Fields`) and `tracing` package has been removed.

Please rebase the changes.

interceptors/tracing/interceptors.go
```diff
@@ -0,0 +1,62 @@
...
+const (
+	SpanKindServer SpanKind = "server"
+	SpanKindClient SpanKind = "client"
+)
+
+type reportable struct {
+	tracer Tracer
+}
+
+func (r *reportable) ServerReporter(ctx context.Context, _ interface{}, typ interceptors.GRPCType, service string, method string) (interceptors.Reporter, context.Context) {
```
yashrsharma44, line 25: You can use the [`CallMeta`](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/v2/interceptors/reporter.go#L58) struct for embedding server/client related info, and this way you can also reduce redundant code.

Here is an example of an implementation - https://github.com/grpc-ecosystem/go-grpc-middleware/blob/v2/interceptors/logging/interceptors.go#L72

interceptors/tracing/reporter.go
```diff
@@ -0,0 +1,68 @@
...
+		o.span.SetAttributes(statusCodeAttr(codes.OK))
+	}
+	o.span.End()
+}
+
+func (o *reporter) PostMsgSend(payload interface{}, err error, d time.Duration) {
+	o.sentMessageID++
+
+	addEvent(o.span, RPCMessageTypeSent, o.sentMessageID, payload)
+}
```
yashrsharma44, line 42:
```suggestion
	o.sentMessageID++
	addEvent(o.span, RPCMessageTypeSent, o.sentMessageID, payload)
}
```

interceptors/tracing/reporter.go
```diff
@@ -0,0 +1,68 @@
...
+
+func (o *reporter) PostMsgReceive(payload interface{}, err error, d time.Duration) {
+	o.receivedMessageID++
+
+	addEvent(o.span, RPCMessageTypeReceived, o.receivedMessageID, payload)
+}
+
```
yashrsharma44, line 49:
```suggestion
func (o *reporter) PostMsgReceive(payload interface{}, err error, d time.Duration) {
	o.receivedMessageID++
	addEvent(o.span, RPCMessageTypeReceived, o.receivedMessageID, payload)
}

```

yashrsharma44 (commented):
Looks good to me, I will have another round of review, particularly the kv package, seems like a lot of chunk for tracing is present there :)

interceptors/tracing/kv/value.go
```diff
@@ -0,0 +1,174 @@
...
+
+type Key string
+
+// Bool creates a KeyValue instance with a BOOL Value.
+//
+// If creating both key and a bool value at the same time, then
+// instead of calling kv.Key(name).Bool(value) consider using a
+// convenience function provided by the api/key package -
+// key.Bool(name, value).
+func (k Key) Bool(v bool) KeyValue {
```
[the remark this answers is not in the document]
XSAM: OTel uses [a type-safe way](https://github.com/open-telemetry/opentelemetry-go/blob/e209ee7585bcb307539a4d58ff16d39b640f58f4/attribute/kv.go#L35) to describe the type of attributes, I am afraid we will lose type information if we don't implement a type system.

interceptors/tracing/interceptors.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+type SpanKind string
+
+const (
+	SpanKindServer SpanKind = "server"
+	SpanKindClient SpanKind = "client"
+)
+
+func reportable(tracer Tracer) interceptors.CommonReportableFunc {
```
yashrsharma44, line 21: This looks neat!

interceptors/tracing/interceptors_test.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+
+	t.spanStore[span.spanID] = &span
+
+	// ctx = contextWithSpan(ctx, &span)
+	if kind == tracing.SpanKindClient {
+		ctx = injectWithContext(ctx, &span, kind)
+	}
+	return ctx, &span
+}
```
yashrsharma44, line 124:
```suggestion
	parentSpan := extractFromContext(ctx, kind)
	if parentSpan != nil {
		// Fetch span from context as parent span
		span.traceID = parentSpan.traceID
		span.parentSpanID = parentSpan.spanID
	} else {
		span.traceID = genID()
	}

	t.spanStore[span.spanID] = &span
	if kind == tracing.SpanKindClient {
		ctx = injectWithContext(ctx, &span, kind)
	}
	return ctx, &span
}
```

leftover comments?

yashrsharma44 (approved):
Everything looks good to me! Regarding the `kv` package, I don't have any solid suggestions, but do you think we can define the interface for the `kv` helper methods, and let the providers implement the `kv` implementation?

XSAM:
> do you think we can define the interface for the kv helper methods, and let the providers implement the kv implementation.

Sounds good. @bwplotka WDYT?

bwplotka:
Just doing pass on this, sorry for lag. Feel free to ping me on Go/CNCF slack next time. Do you wish to move forward with this @XSAM

bwplotka:
Sweet, let me review this unless you want to change something

interceptors/tracing/interceptors_test.go
```diff
@@ -0,0 +1,317 @@
...
+func (s *mockSpan) End() {
+	s.end = true
+}
+
+func (s *mockSpan) SetStatus(code codes.Code, message string) {
+	s.statusCode = code
+	s.statusMessage = message
+}
+
+func (s *mockSpan) AddEvent(name string, attrs ...kv.KeyValue) {
```
bwplotka, line 155:
```suggestion
func (s *mockSpan) AddEvent(name string, attrs ...any) {
```
How bad would be to operate on interfaces (any) and expert implementation to deal with types?

NOTE: I know we could use generics, but in practice they have still some overhead, plus not everyone moved to Go 1.18, so we can add generic type of implementation later on (maybe even in non braking compatibility manner)

bwplotka (commented):
Nice, some comment first. 

Good work 💪🏽

interceptors/tracing/interceptors_test.go
bwplotka: If we would move to interfaces, we won't be needing `kv` package - Otel implementation can then use `"go.opentelemetry.io/otel/attribute"`

interceptors/tracing/kv.go
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) The go-grpc-middleware Authors.
+// Licensed under the Apache License 2.0.
+
+package tracing
+
+type Key string
+
+// KeyValue holds a key and value pair.
+type KeyValue struct {
```
bwplotka, line 9: Why not just:

```
// KeyValue holds a key and value pair.
type KeyValue struct {
	Key   string
	Value interface{}
}
```

bwplotka (approved):
This looks amzing, LGTM. Just one comment. Do you have the OTLP side handy somewhere?

interceptors/tracing/tracing.go
```diff
@@ -0,0 +1,58 @@
...
+	End()
+
+	// SetStatus sets the status of the span in the form of a code
+	// and a message. SetStatus overrides the value of previous
+	// calls to SetStatus on the Span.
+	//
+	// The default span status is OK, so it is not necessary to
+	// explicitly set an OK status on successful Spans unless it
+	// is to add an OK message or to override a previous status on the Span.
+	SetStatus(code codes.Code, msg string)
```
bwplotka, line 48: Amazing idea to use gRPC codes here, love it!

interceptors/tracing/interceptors_test.go
```diff
@@ -0,0 +1,313 @@
...
+func (s *mockSpan) End() {
+	s.end = true
+}
+
+func (s *mockSpan) SetStatus(code codes.Code, message string) {
+	s.statusCode = code
+	s.statusMessage = message
+}
+
+func (s *mockSpan) AddEvent(name string, attrs ...tracing.KeyValue) {
```
bwplotka, line 154: Idea: Remove kv and use `keyvals ...any` as in go-kit logger https://github.com/go-kit/kit/tree/master/log ? (:

bwplotka: well `..interface{}`, since we can't just support Go 1.18 yet

bwplotka: Then we can use something like 

```
// borrowed from https://github.com/go-logfmt/logfmt/blob/main/encode.go#L75
func kvToAttr(keyvals ...interface{}) []attribute.KeyValue {
	if len(keyvals) == 0 {
		return nil
	}
	if len(keyvals)%2 == 1 {
		keyvals = append(keyvals, nil)
	}

	j := 0
	ret := make([]attribute.KeyValue, len(keyvals)/2)
	for i := 0; i < len(keyvals); i += 2 {
		k, ok := anyToString(keyvals[i])
		if !ok {
			k = "<unsupported key type>"
		}
		
		v, ok := anyToString(keyvals[i+1])
		if !ok {
			v = "<unsupported value type>"
		}
		
		ret[j] = attribute.String(k, v)
		j++
	}
	return ret
}

func anyToString(value interface{}) (string, bool) {
	switch v := value.(type) {
	case nil:
		return "nil", true
	case string:
		return v, true
	case []byte:
		return string(v), true
	case error:
		return v.Error(), true
	case fmt.Stringer:
		return v.String(), true
	default:
		rvalue := reflect.ValueOf(value)
		switch rvalue.Kind() {
		case reflect.Array, reflect.Chan, reflect.Func, reflect.Map, reflect.Slice, reflect.Struct:
			return "", false
		case reflect.Ptr:
			if rvalue.IsNil() {
				return "nil", true
			}
			return anyToString(rvalue.Elem().Interface())
		}
		return fmt.Sprint(v), true
	}
}

```

bwplotka:
Ping (:

bwplotka (approved):
Amazing, work thanks! Let's go!

PR #459, opened 2021-09-17
Add a method filter for interceptors

CAFxX:
This PR adds a conditional interceptor for specific gRPC methods.

It allows to apply an interceptor (chain) only to specific methods, or to all but some specific methods.

### DIFF
--- a/filter/server_interceptors.go
+++ b/filter/server_interceptors.go
@@ -1,0 +1,65 @@
+package filter
+
+import (
+	"context"
+
+	"google.golang.org/grpc"
+)
+
+// UnaryServerMethods returns an interceptor that applies the provided interceptor only to incoming unary calls to the specified methods.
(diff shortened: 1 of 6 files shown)


filter/server_interceptors.go
```diff
@@ -0,0 +1,65 @@
...
+		panic("nil interceptor")
+	}
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		if condition(ctx, req, info) {
+			return interceptor(ctx, req, info, handler)
+		}
+		return handler(ctx, req)
+	}
+}
+*/
```
CAFxX, line 47: this is an alternate implementation style in case we want to support plugging in arbitrary conditions

Issue #460, opened 2021-09-20
feat: Adding documentation website for Go gRPC Middleware


iamrajiv:
## Description

We can have a docs website for Go gRPC Middleware same as [https://grpc-ecosystem.github.io/grpc-gateway/](https://grpc-ecosystem.github.io/grpc-gateway/).

iamrajiv:
@bwplotka @yashrsharma44 Can I work on this?

yashrsharma44:
How much effort would be to create one? Would be awesome to have if, but not a blocker for v2.



> @bwplotka @yashrsharma44 Can I work on this?

Sure let me assign you 😀

iamrajiv:
@yashrsharma44 We can replicate the same theme that we have for gRPC-Gateway. First I will initialize the docs site with the theme and small content. In future PRs we can start adding content. I don't think it will be a blocker for `v2` like this will not affect any codebase because I will be adding a new `docs` folder and inside that, all docs site configuration and the content will be there.

yashrsharma44:
Sure, go ahead with the initial work, let me know if you need help, we can figure out to break the tasks.

PR #461, opened 2021-09-26
feat: Initialized Go gRPC Middleware docs site

iamrajiv:
## Description

Fixes: #460

- [x] Initialized the docs site with [Just the Docs](https://github.com/pmarsceill/just-the-docs) Jekyll theme.
- [x] Refactored README.


iamrajiv:
Hi, @bwplotka @johanbrandhorst @yashrsharma44, I have initialized the docs. Also, we need to add a new GitHub token to the repository with the name `JEKYLL_GITHUB_TOKEN` because we need it to build the site automatically whenever there is a change in the docs folder.

bwplotka:
Hi, 

Thank you for this contribution, but I think we would love to choose a technology that is closer to what maintainers know about. The reason is that we need to maintain it over time. But generally, the idea to have a website is amazing! We recently used https://github.com/bwplotka/mdox and Hugo for this job, so if you want to help, let's aim for that instead of jekyll & co. 

You can learn about mdox here https://www.bwplotka.dev/2021/mdox

bwplotka:
Also this has to be on `v2` if we still wants this.

bwplotka:
Closing, unless you want to continue on this, we can then reopen.

PR #414, opened 2021-04-21
wrong example misguide user to catch panic

Wander1024:
Signed-off-by: wander1024 <yawhua1024@gmail.com>

recovery should be placed the first.

#410 
#215

PR #464, opened 2021-10-18
Doc Fix: middleware chain order example

AmirQadir:
In the chaining interceptor example, the recovery handler was the first element in the unary chain. It should have been last as per the recovery example in the codebase.

https://github.com/grpc-ecosystem/go-grpc-middleware/blob/b6d97fab8b7d7d99cb5a9ee1ea17afbb2b12e84c/recovery/examples_test.go#L28
https://github.com/grpc-ecosystem/go-grpc-middleware/blob/b6d97fab8b7d7d99cb5a9ee1ea17afbb2b12e84c/recovery/examples_test.go#L29

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,20 +29,20 @@ import "github.com/grpc-ecosystem/go-grpc-middleware"
 
 myServer := grpc.NewServer(
     grpc.StreamInterceptor(grpc_middleware.ChainStreamServer(
-        grpc_recovery.StreamServerInterceptor(),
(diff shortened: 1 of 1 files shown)


xiaozhoz:
I intend to put it to the first one since it can recovery unexpected panic from other custom middlewares. see https://github.com/grpc-ecosystem/go-grpc-middleware/pull/414 also
I didn't get it how it affects other middlewares if recovery is the first one. Could you please explain more?

PR #469, opened 2021-11-03
Add factory based zap payload logging interceptor

deepakgupta33:
Feature Request:
We have a use case to extract key-value pairs from the context for the payload logging.

Current state:
The context sent by the grpc framework to the interceptor but there is no hook for payload interceptor user to plug into and be able to decorate the logger with the information in the context object. There are other interceptors which populate the context object.

Solution:
Instead of directly passing on the logger, we create another method that accepts a logger factory. The logger is created only after the context is received with the context object.


deepakgupta33:
@googlebot signed it!

PR #472, opened 2021-11-17
Replacing deprecated marshaller in the interceptors

rodoufu:
The tests are failing cause the generated proto files are using the deprecated libraries, how can I fix that?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/grpc-ecosystem/go-grpc-middleware
 require (
 	github.com/go-kit/log v0.1.0
 	github.com/gogo/protobuf v1.3.2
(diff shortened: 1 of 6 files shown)


johanbrandhorst:
I don't think we can make this upgrade on `master` - it's one of the reasons we started working on a `v2`. Can you try using the `v2` branch instead?

rodoufu:
> I don't think we can make this upgrade on `master` - it's one of the reasons we started working on a `v2`. Can you try using the `v2` branch instead?

I will, thanks @johanbrandhorst

rodoufu:
It seems to have already been solved in `v2` so I'm closing this PR.

PR #476, opened 2021-11-29
Extract zap logger from context for logging server interceptors

implmnt:
This PR extends the current zap logging server interceptors API. 
It makes possible to reuse existing logger from GRPC request context for logging user's contextual data (e.g. the `user_id` that was shifted before into the logging context).

### DIFF
--- a/logging/zap/server_interceptors.go
+++ b/logging/zap/server_interceptors.go
@@ -20,13 +20,23 @@ var (
 	ServerField = zap.String("span.kind", "server")
 )
 
+type extractLogger = func(ctx context.Context) *zap.Logger
+
 // UnaryServerInterceptor returns a new unary server interceptors that adds zap.Logger to the context.
 func UnaryServerInterceptor(logger *zap.Logger, opts ...Option) grpc.UnaryServerInterceptor {
+	return LoggingUnaryServerInterceptor(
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
Hi Georgy. I'm not sure this is a change we want to make, could you explain the rationale for this new API? Please note that we have moved all new feature development to the v2 branch too, where we have a new structure for these sort of implementations.

implmnt:
Hi Johan. Thanks for notifying me about v2 branch. Will check it too.

The reason of the new API is making possible to reuse logger that was inited in some before step.

For expample:
I have an interceptor which lifts `user_id` to logger context and I expect to see it through all logs of whole application.
The current logging interceptor allows to use the only logger that was passed through its context-less constructor and does not provide access to the request context.
`func UnaryServerInterceptor(logger *zap.Logger, opts ...Option) grpc.UnaryServerInterceptor {`

So I think it would be better to have a more flexible API that allows extracting logger from request context with some predefined fields like `user_id`.

johanbrandhorst:
@bwplotka do you have any thoughts on this?

bwplotka:
Sorry for lag!

Yes, while this is great, we have opportunity to change things on `v2` - we would love to not add new features to current v1 (master)

PR #490, opened 2022-04-01
Update provider/kit

metalmatze:
There were incompatibilities when requiring the latest commit for the main module. 
Hence I've added the ```
replace github.com/grpc-ecosystem/go-grpc-middleware/v2 => ../..```
similar to what the provider/openmetrics does. 

Now the only problem is that I'm not 100% sure how the removal of the tags package is supposed to work here. Additionally, the `payloadDecider` needs to be fixed too. I wasn't able to find exactly which type implements that.

/cc @bwplotka


bwplotka (approved):
Thanks 👍🏽

bwplotka:
Plus linter is failing

metalmatze:
I'm not sure if I correctly removed the `tags` package if there was something I missed or it really was just about deleting the lines in this PR?

bwplotka:
I don't know, I would need to look into the details. This PR looks like a good direction at least 🙃

PR #495, opened 2022-06-08
Add opentelemetry provider

XSAM:
Split from https://github.com/grpc-ecosystem/go-grpc-middleware/pull/312.

@bwplotka @yashrsharma44 PTAL


johanbrandhorst:
Hi @XSAM, thank you for the PR. I'm a little confused, why do we need an opentelemetry provider when `otelgrpc` exists? I know this has been in the works for ages, but I'd like to encourage users _not_ to use us when there are perfectly good alternatives.

XSAM:
Hi @johanbrandhorst, My original thought is we can control the implementation of the opentelemetry provider, like adding more custom attributes. But I am also a little confused about whether this feature is actually needed.

Feel free to close this PR if there are no other reasons to have this provider.

johanbrandhorst:
Lets wait until we find a compelling use case, thanks!

bwplotka:
Hmmm,  I wonder if ease of use is one of those use cases. Currently the otel libraries are not the easiest to import and use. (e.g. I wrote wrapper for same reason: https://github.com/bwplotka/tracing-go

Having one, consistent way of doing things might helpful. Maybe indeed we could wait for specific use case.

Issue #498, opened 2022-07-07
kit.PayloadUnaryServerInterceptor memory leakage


r3code:
During work of an app, this code consumes memory without returning it, and it can cause Out of memory one day
https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/logging/kit/payload_interceptors.go#L34

karelbilek:
This issue can be closed I think; it's merged into master

PR #501, opened 2022-07-12
Fix memory leakage in kit.PayloadUnaryServerInterceptor

r3code:
Closes issue #498

A wrong assign creates multiple object for GC, and with every call appends new objects, so GC have to walk through thousands of objects. This causes high increasing memory and CPU usage. We saw in our services that it leads to CPU throttling and increasing of a memory usage.


johanbrandhorst (approved):

johanbrandhorst:
Thank you for this contribution!

PR #502, opened 2022-07-17
Use %+v to store the error

apratina:
Using `%+v` to store the error allows to print stacktrace along with the error message, if errors are constructed using the `github.com/pkg/errors package`. See https://dave.cheney.net/2016/06/12/stack-traces-and-the-errors-package for more details.

If error is created using `errors` package, it works as is and has no impact in behavior.

Maintainers - Let me know if you have some other suggestions.

### DIFF
--- a/logging/logrus/options.go
+++ b/logging/logrus/options.go
@@ -5,6 +5,7 @@ package grpc_logrus
 
 import (
 	"context"
+	"fmt"
 	"time"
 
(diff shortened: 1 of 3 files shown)


johanbrandhorst (commented):
This destroys error information, it doesn't preserve it. If there's any error stack information we want to preserve, it's that which is contained within the error itself, printing it will remove this.

PR #503, opened 2022-07-17
Use strconv.FormatUint instead of fmt.Sprintf

berenda-andrey:
strconv.FormatUint a little bit faster that fmt.Sprintf


berenda-andrey:
@googlebot signed it!

johanbrandhorst (approved):
LGTM

PR #507, opened 2022-07-22
Add optional log level for payload interceptors.

slavaromanov:


johanbrandhorst (commented):
As crazy as this may sound, this is actually a breaking change, so I don't think we can add this in v1. It's breaking because if someone has assigned the value of the existing function to a variable, it will change the type of that variable with this change. Would you like to contribute this to the v2 interceptor instead?

slavaromanov:
OK, I'll do a PR in v2 later.

Issue #391, opened 2021-01-30
v2: Switch `grpclog.Infof` with something non-global


bwplotka:
There is no way to control this on instance lvl, it's only global. Let's fix this.

e.g `grpclog.Infof("grpc_opentracing: failed parsing trace information: %v", err)` in `tracing` package.

Not blocking v2 release, we can do later.

berenda-andrey:
@bwplotka I have created the pr
could you watch it?

bwplotka:
Thanks for the PR, but this is now done, because we decided to remove opentracing completely. It is deprecated (if needed opentracing from v1 can be used) and otherwise users can use otlpgrpc (: 

Thanks anyway, and sorry we didn't notice this earlier!

PR #508, opened 2022-07-27
introduce errorLogFunc in opentracing options

andrey-berenda:
Issue #391


johanbrandhorst:
Hi, I believe we want this functionality against the `v2` branch, not `master`. Could you try to rewrite this against that branch? CC @bwplotka

bwplotka:
Amazing work, this is how we would do it.. if the issue would be still relevant. 

Mentioned details in https://github.com/grpc-ecosystem/go-grpc-middleware/issues/391

Thanks anyway, great job! 🤗

PR #511, opened 2022-09-02
✨ new feat: selector middleware

aimuz:
It allows to set check rules to whitelist or blacklist middleware such as Auth
Requests that match the rules will go to the next level of middleware

### DIFF
--- a/interceptors/selector/doc.go
+++ b/interceptors/selector/doc.go
@@ -1,0 +1,15 @@
+// Copyright (c) The go-grpc-middleware Authors.
+// Licensed under the Apache License 2.0.
+
+/*
+Package selector
+
+`selector` a generic server-side selector middleware for gRPC.
+
+# Server Side Selector Middleware
+It allows to set check rules to allowlist or blocklist middleware such as Auth
+interceptors to toggle behavior on or off based on the request path.
+
+Please see examples for simple examples of use.
+*/
+package selector
--- a/interceptors/selector/selector.go
+++ b/interceptors/selector/selector.go
@@ -1,0 +1,34 @@
+// Copyright (c) The go-grpc-middleware Authors.
+// Licensed under the Apache License 2.0.
+
+package selector
+
+import (
+	"context"
+
+	"google.golang.org/grpc"
+)
+
+type MatchFunc func(ctx context.Context, fullMethod string) bool
+
+// UnaryServerInterceptor returns a new unary server interceptor that will decide whether to call
+// the interceptor on the behavior of the MatchFunc.
+func UnaryServerInterceptor(interceptors grpc.UnaryServerInterceptor, match MatchFunc) grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{}, err error) {
+		if match(ctx, info.FullMethod) {
+			return interceptors(ctx, req, info, handler)
+		}
+		return handler(ctx, req)
+	}
+}
+
+// StreamServerInterceptor returns a new stream server interceptor that will decide whether to call
+// the interceptor on the behavior of the MatchFunc.
+func StreamServerInterceptor(interceptors grpc.StreamServerInterceptor, match MatchFunc) grpc.StreamServerInterceptor {
+	return func(srv interface{}, ss grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
+		if match(ss.Context(), info.FullMethod) {
+			return interceptors(srv, ss, info, handler)
+		}
+		return handler(srv, ss)
+	}
+}
--- a/interceptors/selector/selector_example_test.go
+++ b/interceptors/selector/selector_example_test.go
@@ -1,0 +1,84 @@
+// Copyright (c) The go-grpc-middleware Authors.
+// Licensed under the Apache License 2.0.
+
+package selector_test
+
+import (
+	"context"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/auth"
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/logging"
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/ratelimit"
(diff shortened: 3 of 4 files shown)


johanbrandhorst (changes requested):
This looks cool, but I think it's missing some things:

1. Where are the tests? We should include some tests of this new functionality.
2. Could you include an `selector_example_test.go` file with an example of how to use this? It will automatically be shown as part of the documentation. See https://go.dev/blog/examples.

aimuz:
Yes, that's how it should be. I'll take care of it.

interceptors/selector/doc.go
```diff
@@ -0,0 +1,12 @@
+/*
+Package selector
+
+`selector` a generic server-side selector middleware for gRPC.
+
+# Server Side Selector Middleware
+It allows to set check rules to whitelist or blacklist middleware such as Auth
```
johanbrandhorst, line 7: Thanks for the updates, do you mind changing these phrases to `allowlist` and `blocklist` consistently through this PR?
```suggestion
It allows to set check rules to allowlist or blocklist middleware such as Auth
```

interceptors/selector/selector.go
```diff
@@ -0,0 +1,58 @@
...
+
+type MatchFunc func(ctx context.Context, fullMethod string) bool
+
+type options struct {
+	match MatchFunc
+}
+
+type Option func(o *options)
+
+// WithMatch ...
```
johanbrandhorst, line 17: I think this comment needs improving
```suggestion
// WithMatch configures what condition should trigger the selector to run the interceptor.
// If used multiple times, the last one specified will be used.
```

interceptors/selector/selector.go
```diff
@@ -0,0 +1,58 @@
...
+func evaluateServerOpt(opts []Option) *options {
+	optCopy := &options{}
+	optCopy.match = AlwaysMatch
+	for _, o := range opts {
+		o(optCopy)
+	}
+	return optCopy
+}
+
+// AlwaysMatch is the default implementation, always selected
```
johanbrandhorst, line 33:
```suggestion
// AlwaysMatch can be used to always run the interceptor. This is the default behavior.
```

interceptors/selector/selector.go
```diff
@@ -0,0 +1,58 @@
...
+func AlwaysMatch(_ context.Context, _ string) bool {
+	return true
+}
+
+// UnaryServerInterceptor returns a new unary server interceptors that performs per-request selector.
```
johanbrandhorst, line 38:
```suggestion
// UnaryServerInterceptor returns a new unary server interceptor that will decide whether to call the interceptor on the behavior of the MatchFunc.
```

interceptors/selector/selector.go
```diff
@@ -0,0 +1,58 @@
...
+func UnaryServerInterceptor(interceptors grpc.UnaryServerInterceptor, opts ...Option) grpc.UnaryServerInterceptor {
```
johanbrandhorst, line 39: I'm leaning towards removing `MatchFunc` as an option and making it a required parameter. This solves two problems:

1. If you accidentally use the selector without a match func, you gain no utility from it.
2. If you accidentally specify two match func options, you might be confused about which to use.

That simplifies things generally, no need for the `AlwaysMatch` function, etc.

interceptors/selector/doc.go
```diff
@@ -0,0 +1,12 @@
...
+Requests that match the rules will go to the next level of middleware
```
johanbrandhorst, line 8:
```suggestion
interceptors to toggle behavior on or off based on the request path.
```

interceptors/selector/selector_example_test.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	ctx = logging.InjectFields(ctx, logging.Fields{"auth.sub", userClaimFromToken(tokenInfo)})
+
+	// WARNING: In production define your own type to avoid context collisions.
+	return context.WithValue(ctx, tokenInfoKey, tokenInfo), nil
+}
+
+func loginSkip(ctx context.Context, fullMethod string) bool {
+	return fullMethod != "login"
```
johanbrandhorst, line 67: Lets use a realistic looking path, like you are doing in the tests
```suggestion
	return fullMethod != "/auth.v1.AuthService/Login"
```

johanbrandhorst (changes requested):
Thanks for the tests, just a few suggestions

interceptors/selector/selector.go
aimuz: Yes, I readjusted it, can you help me in seeing what is wrong?

interceptors/selector/selector.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+}
+
+// Block the interceptor will not run after the method matches
+func Block(methods []string) MatchFunc {
+	allow := Allow(methods)
+	return func(ctx context.Context, fullMethod string) bool {
+		return !allow(ctx, fullMethod)
+	}
+}
```
johanbrandhorst, line 29: I don't think we need these, to be honest. Users can easily implement these themselves.

interceptors/selector/selector_test.go
```diff
@@ -0,0 +1,132 @@
...
+import (
+	"context"
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/auth"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+	"google.golang.org/grpc"
+	"testing"
+)
+
+var whiteList = []string{"/auth.v1beta1.AuthService/Login"}
```
johanbrandhorst, line 12:
```suggestion
var allowList = []string{"/auth.v1beta1.AuthService/Login"}
```

johanbrandhorst (changes requested):

interceptors/selector/selector_test.go
aimuz: I have adjusted

testing/testpb/test.pb.go
```diff
@@ -1,18 +1,16 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.25.0
+// 	protoc-gen-go v1.28.0
```
johanbrandhorst, line 3: Can you revert these changes, I don't think these should need to be changed?

johanbrandhorst (changes requested):

johanbrandhorst (approved):

johanbrandhorst:
Looks like there is a formatting issue, can you run `make lint`?

aimuz:
done

PR #512, opened 2022-09-08
change the doc.go to the latest format

aimuz:
This is the result after formatting with gofmt and minor modifications

### DIFF
--- a/providers/kit/doc.go
+++ b/providers/kit/doc.go
@@ -2,7 +2,7 @@
 // Licensed under the Apache License 2.0.
 
 /*
-`kit` provides a small adapter required to use go-kit/log in logging gRPC middlewares.
- Please see examples for examples of use.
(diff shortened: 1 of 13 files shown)


providers/kit/doc.go
```diff
@@ -3,6 +3,7 @@
 
 /*
 `kit` provides a small adapter required to use go-kit/log in logging gRPC middlewares.
- Please see examples for examples of use.
+
+	Please see examples for examples of use.
```
johanbrandhorst, line 7: I think this and the other three instances of this change is wrong, can you revert them, please?

johanbrandhorst (changes requested):

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

PR #522, opened 2022-12-12
use idiomatic pointer safing for zap.Check

ermik:
Call to zap.Logger Check method is a proxy to internal [check][1] which can return nil. Looking at the way Uber's own code, they use the idiomatic pointer safing before continuing past that call.

If accepted, this commit should be a bug fix to all actively maintained versions.

[1]: https://github.com/uber-go/zap/blob/v1.18.1/logger.go#L261


johanbrandhorst:
This doesn't change the behavior, the `CheckedEntry` type already returns early if called on a `nil` pointer: https://github.com/uber-go/zap/blob/v1.18.1/zapcore/entry.go#L202. Thanks for checking in though!

ermik:
Ah, smart. I totally forgot that calling a method on a nil type wasn't an immediate death sentence. I am stumped though on what the nil pointer dereference could be on that line. I don't have a reliable reproduction, but this was my best guess and it seemed to work. Appreciate the review and sorry for bugging you! <3

Issue #515, opened 2022-09-25
[ERROR]: ambiguous import in auth/auth_test.go


panjf2000:
I ran `go mod tidy` and get this output: 

```powershell
	github.com/grpc-ecosystem/go-grpc-middleware/auth tested by
	github.com/grpc-ecosystem/go-grpc-middleware/auth.test imports
	google.golang.org/grpc/credentials/oauth imports
	golang.org/x/oauth2/google imports
	cloud.google.com/go/compute/metadata: ambiguous import: found package cloud.google.com/go/compute/metadata in multiple modules:
	cloud.google.com/go v0.65.0 (/Users/andypan/dev/go/pkg/mod/cloud.google.com/go@v0.65.0/compute/metadata)
	cloud.google.com/go/compute v1.7.0 (/Users/andypan/dev/go/pkg/mod/cloud.google.com/go/compute@v1.7.0/metadata)
```

Related issue: #377

johanbrandhorst:
Which branch was this? I can run this just fine on `master`. Do you need to update one of your dependencies?

panjf2000:
> Which branch was this? I can run this just fine on `master`. Do you need to update one of your dependencies?

v1.3.0, and it can be resolved by adding `eplace cloud.google.com/go v0.65.0 => cloud.google.com/go/compute v1.7.0
` to go.mod

johanbrandhorst:
Thanks for the workaround. Do you think we need to update our dependency?

panjf2000:
Hi @johanbrandhorst , sorry for the delay, it's been a busy month. The ambiguous import is caused by [oauth2](https://github.com/golang/oauth2), and `cloud.google.com/go v0.65.0` in go.mod of that repo has been replaced to `cloud.google.com/go/compute v1.7.0`, see [this commit](https://github.com/golang/oauth2/commit/2104d58473e0869996f17ef05fb7e0cc9bd19312),  thus the solution for this issue is to bump up the dependencies of go-grpc-middleware.

[...]

PR #527, opened 2023-01-06
Bumps oauth2 dependencies

nickzelei:
The commands I ran:

```
go get golang.org/x/oauth2
go get golang.org/x/oauth2/google@v0.4.0
go mod tidy
```

Not sure if this is going to be right, but I'm trying to fix the problem in #515 
I want to use the grpc_auth middleware, but can't due to outdated dependencies.


nickzelei:
Is there a way I can get my github noreply email added to my CLA? I couldn't figure out how to do so. (Individual contributor)

johanbrandhorst:
Hm, I'm not sure, I've only used it with my github email.

bwplotka:
This is now fixed. Thanks anyway!

PR #535, opened 2023-02-17
Add JWT support to the auth middleware

ErenDursun:
Adds JWT support to the auth middleware.

For example verifying JWT via 'HMAC' can be done like this:
```
import grpc_jwt "github.com/grpc-ecosystem/go-grpc-middleware/auth/jwt"

jwtAuthFunc := grpc_jwt.NewAuthFunc([]byte("some_secret"))
svr := grpc.NewServer(
	grpc.StreamInterceptor(grpc_auth.StreamServerInterceptor(jwtAuthFunc)),
	grpc.UnaryInterceptor(grpc_auth.UnaryServerInterceptor(jwtAuthFunc)),
)
```

If you want to use 'ES256' it would look like this:
```
import grpc_jwt "github.com/grpc-ecosystem/go-grpc-middleware/auth/jwt"

key, _ := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
jwtAuthFunc := grpc_jwt.NewAuthFuncWithConfig(
	grpc_jwt.Config{
		SigningMethod: jwt.SigningMethodES256.Name,
		SigningKey:    key.PublicKey,
	},
)
svr := grpc.NewServer(
	grpc.StreamInterceptor(grpc_auth.StreamServerInterceptor(jwtAuthFunc)),
	grpc.UnaryInterceptor(grpc_auth.UnaryServerInterceptor(jwtAuthFunc)),
)
```

The implementation is heavily based on [labstack/echo-jwt](https://github.com/labstack/echo-jwt).
The cryptographic algorithms are implemented by [golang-jwt/jwt](https://github.com/golang-jwt/jwt) (currently: ECDSA, Ed25519, HMAC, RSA, RSAPSS).

Also, I might need some help adding more documentation on how to use. Currently there's only [this](https://github.com/ErenDursun/go-grpc-middleware/blob/45192cace81b26c58684ec1e203051e98b535752/auth/jwt/jwt.go#L16) and [this](https://github.com/ErenDursun/go-grpc-middleware/blob/45192cace81b26c58684ec1e203051e98b535752/auth/jwt/examples_test.go#L27).


johanbrandhorst:
Hi, thanks for your PR. I'm sorry that you've put in all this effort to add this and I'm going to have to say that we don't want this. I'll explain:

1. It makes me nervous that we would implement some generic JWT authentication solution. Authentication is a very _very_ important part of any application, and I don't think it's our place to do this work in a generic handler like this. I would hate to b responsible for someone implementing authentication using this without understanding what they are doing.
2. This implementation targets the `master` branch, which we don't want to add new features to, this would have to go through the v2 branch.

We could have avoided all this wasted work if you created an issue to ask whether we wanted this feature first. Having said all of that, it might be cool to have this implementation as an example of the use of our existing authentication middleware.

ErenDursun:
Hi @johanbrandhorst , thanks for the quick reply. Totally missed that there was a v2 branch. Although adapting the implementation for v2 would be quite easy. But if adding a generic JWT auth handler is outside the scope of this project, I think I'll just setup a separate repo using go-grpc-middleware as a third-party dependency.
Anyways, thanks for this awesome project! :) Just wanted to give something back and contribute 😄 

I'll add a link here when ready.

ErenDursun:
Hope it can be of use to someone 😃
[ErenDursun/go-grpc-jwt-middleware](https://github.com/ErenDursun/go-grpc-jwt-middleware)

fckveza:
Very useful for me.. thank you @ErenDursun 😊

Issue #483, opened 2022-02-15
`grpc_retry.StreamClientInterceptor` does not pass the correct `"x-retry-attempty"` metadata value


slnt:
In `retry.go` in the implementation of `StreamClientInterceptor` there is the following code:

```go
for attempt := uint(0); attempt < callOpts.max; attempt++ {
if err := waitRetryBackoff(attempt, parentCtx, callOpts); err != nil {
	return nil, err
}
callCtx := perCallContext(parentCtx, callOpts, 0)
```

This code _should_ be:

```go
for attempt := uint(0); attempt < callOpts.max; attempt++ {
if err := waitRetryBackoff(attempt, parentCtx, callOpts); err != nil {
	return nil, err
}
callCtx := perCallContext(parentCtx, callOpts, attempt)
```

Also, there seems to be an extraneous `y` character in the definition of `AttemptMetadataKey`:
```go
const (
	AttemptMetadataKey = "x-retry-attempty"
)
```

johanbrandhorst:
Hi, thanks for the issue. It appears at least the `AttemptMetadataKey` typo is also present in v2, though the major issue here does appear to be addressed: https://github.com/grpc-ecosystem/go-grpc-middleware/blob/a5b9e0b0458cc2e39faab057564de9efe962f0ba/interceptors/retry/retry.go#L30-L70.

bwplotka:
Typo fixed in https://github.com/grpc-ecosystem/go-grpc-middleware/pull/543

Issue #422, opened 2021-05-18
Exemplars support: Add ability to include trace ID


lilic:
Hello 👋 

Is there any way currently (I haven't been able to find it yet) to include the Trace ID as part of the interceptor to be able to make use of Prometheus Exemplars? If not, are there plans to add this? 

Thank you and sorry if I missed something obvious!
@bwplotka @brancz

bwplotka:
There is no request yet, but thanks for raising it, definitely help wanted on that.

We are planning to add exemplars to HTTP middleware too: https://github.com/prometheus/client_golang/issues/854 so Ideally we can follow similar pattern. One point is that Go tracing client lib for OpenTelemetry is not yet stable, so we might want to wait on that, unless we want to abstract that away (probably good idea)

lilic:
Thank you! Is there a need to tie the OpenTel client lib to this repo, or can we make the ID generic somehow like you suggested with abstracting it away?

bwplotka:
I would definitely keep it generic, but would be interesting to mix tracing and metrics libs together (: 

Can we just add tracing to client_prometheus? 🙈 cc @kakkoyun

lilic:
> I would definitely keep it generic, but would be interesting to mix tracing and metrics libs together (:

I fear that would cause go.mod dep issues, so if possible to keep those separate would be ideal.

What are the actual things we need to do, I was under the impression we would "just" need to add a trace ID as label key/value, and that is not specific to OpenTel? But again I can be very wrong, didn't get a chance to deep dive into it.

bwplotka:
Well, the main question is where you get this trace ID from. Then the question is do you want to put trace that is sampled? How do we know that etc. 


We can of course just allow user to specify this but again, some example implementation using Otel, Opentracing would be helpful (see v2, we have multiple modules for this)

brancz:
The way I understand it is that this library should just use the [w3c trace context](https://www.w3.org/TR/trace-context/), which opentelemetry happens to [support](https://github.com/open-telemetry/opentelemetry-dotnet/issues/135), but is ultimately independent of any specific tracing implementation, which seems like the right fit for a library like this I would say.

bwplotka:
Done in https://github.com/grpc-ecosystem/go-grpc-middleware/pull/543

Issue #347, opened 2020-10-02
v2: Add failint to avoid not needed imports (e.g v1 packages)


bwplotka:
Faillint usage example: https://github.com/thanos-io/thanos/blob/master/Makefile#L305

Pulled from https://github.com/grpc-ecosystem/go-grpc-middleware/issues/275 for visibility.

bwplotka:
In progress https://github.com/grpc-ecosystem/go-grpc-middleware/pull/543

Issue #345, opened 2020-10-02
v2 Consider adding allow-list behaviour to auth interceptor


bwplotka:
AC:
* Define simple functionality for auth interceptor before stating stable version: How to enable whitelist auth? Any interface changes?

Pulled from https://github.com/grpc-ecosystem/go-grpc-middleware/issues/275 for visibility.

Blocker for v2.

jzelinskie:
We actually wrote a very simple mixin that you embed in a server the auth interceptor will ignore it -- this is most useful for the healthcheck service: https://pkg.go.dev/github.com/authzed/grpcutil#IgnoreAuthMixin

devnev:
There's also two interceptor packages relating to selectively enabling other interceptors: `skip` and `selector`.

So possible interface could something like one of the following:
```
    auth.UnaryServerInterceptor(authFunc, auth.IgnoreMethods("grpc.health.v1.Health/Check"))
    auth.UnaryServerInterceptor(authFunc, auth.IgnoreServices(health_pb.Health_ServiceDesc)
    selector.UnaryServerInterceptor(auth.UnaryServerInterceptor(authFunc), selector.Not(selector.MatchMethods("grpc.health.v1.Health/Check")))
    selector.UnaryServerInterceptor(auth.UnaryServerInterceptor(authFunc), selector.Not(selector.MatchServices(health_pb.Health_ServiceDesc)))
    skip.UnaryServerInterceptor(auth.UnaryServerInterceptor(authFunc), skip.Methods("grpc.health.v1.Health/Check"))
    skip.UnaryServerInterceptor(auth.UnaryServerInterceptor(authFunc), skip.Services(health_pb.Health_ServiceDesc))
```

This could also be an argument to adjust the interfaces of the `skip` and `selector` packages, or to merge them.

bwplotka:
Great point. I am then considering closing this issue as it's doable with selector and skip... Question is, is this interface nice and if not, can we improve it before v2?

bwplotka:
Actually you just mentioned improvements ideas... 

1. Auth is great, but we need selection for many other interceptors so skip/selector works better.
2. Selector looks nice, it's bit explicit but clean. It also allows skipping "Not" for allow-list.
3. Skip is nice, but negation of skip would look weird (skip.Not)

So 2?

bwplotka:
[...]

PR #543, opened 2023-03-15
v2: All for v2: Exemplars, Cleanup, Docs, Lint, Proto upgrades and more

bwplotka:
Fixes https://github.com/grpc-ecosystem/go-grpc-middleware/issues/347
Fixes https://github.com/grpc-ecosystem/go-grpc-middleware/issues/343
Fixes https://github.com/grpc-ecosystem/go-grpc-middleware/issues/345
Fixes https://github.com/grpc-ecosystem/go-grpc-middleware/issues/483
Fixes https://github.com/grpc-ecosystem/go-grpc-middleware/issues/422

Most (if not all) changes for v2.0.0!

Let me know if I need to split into commits for easier review.

TODO: 
* [x] Finish migration guide
* [x] Finish example README
* [X] Find out how to change main to v2 without breaking v1 users. All tested https://github.com/golang/go/wiki/Modules#faqs--multi-module-repositories . One finding: We will need to have multiple releases for each submodule in providers (will create quick script). We can start with 0.1.0 for some of them too and probably it makes sense to do so. Changing default branch to "v2" once v2.0.0 is out, should be fine.


bwplotka:
At least one thing still bothers me before v2:

So, `interceptors/*` have interceptors that sometimes consumes providers. Ironically providers have literally only logging adapters and prometheus interceptor (!). It might feel quite odd. Perhaps moving providers/prometheus to interceptors/prometheus (as a separate module) might make more sense? I don't think we plan to add any other metric middleware so it could be equally interceptors/metrics 🤔 (also avoids package name collision). The only odd thing about this alternative is that "metrics" will be the only interceptor that is a separate module (if we want to NOT pollute go-grpc-middleware with very common prometheus import)... WDYT @johanbrandhorst @devnev ?

bwplotka:
All done and ready for review.

examples/client/main.go
```diff
@@ -0,0 +1,148 @@
...
+			logging.StreamClientInterceptor(kit.InterceptorLogger(rpcLogger), logging.WithFieldsFromContext(logTraceID))),
+	)
+	if err != nil {
+		level.Error(logger).Log("err", err)
+		os.Exit(1)
+	}
+
+	cl := testpb.NewTestServiceClient(cc)
+
+	g := &run.Group{}
```
johanbrandhorst, line 102: Minor nit, but can't we just use `errgroup`, which is much more widely used than `oklog/run`?

interceptors/auth/examples_test.go
```diff
@@ -7,9 +7,10 @@ import (
 	"context"
 	"log"
 
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/testing/testpb"
```
johanbrandhorst, line 10: Nit: import ordering

interceptors/callmeta.go
```diff
@@ -0,0 +1,66 @@
...
+func splitFullMethodName(fullMethod string) (string, string) {
+	fullMethod = strings.TrimPrefix(fullMethod, "/") // remove leading slash
+	if i := strings.Index(fullMethod, "/"); i >= 0 {
+		return fullMethod[:i], fullMethod[i+1:]
+	}
+	return "unknown", "unknown"
+}
+
+type CallMeta struct {
+	ReqOrNil interface{}
```
johanbrandhorst, line 22: Shall we use `any`? Since we're supporting Go 1.19+ anyway.

interceptors/client_test.go
```diff
@@ -11,6 +11,8 @@ import (
 	"testing"
 	"time"
 
+	"google.golang.org/grpc/credentials/insecure"
```
johanbrandhorst, line 14: Nit: import grouping

interceptors/logging/options.go
```diff
@@ -4,9 +4,12 @@
 package logging
 
 import (
+	"context"
 	"fmt"
 	"time"
 
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors"
```
johanbrandhorst, line 11: I think we want this group last, no?

interceptors/selector/selector.go
```diff
@@ -6,29 +6,76 @@ package selector
 import (
 	"context"
 
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors"
```
johanbrandhorst, line 9: Imports

interceptors/selector/selector.go
```diff
@@ -6,29 +6,76 @@ package selector
...
+		if matcher.Match(ss.Context(), c) {
+			return i(srv, ss, info, handler)
 		}
 		return handler(srv, ss)
 	}
 }
+
+// UnaryClientInterceptor returns a new unary client interceptor that will decide whether to call
+// the interceptor based on the return argument from the Matcher.
+// TODO(bwplotka): Write unit test.
```
johanbrandhorst, line 59: Looks like we can remove this

interceptors/selector/selector.go
```diff
@@ -6,29 +6,76 @@ package selector
...
+		if matcher.Match(ctx, c) {
+			return i(ctx, method, req, reply, cc, invoker, opts...)
+		}
+		return invoker(ctx, method, req, reply, cc, opts...)
+	}
+}
+
+// StreamClientInterceptor returns a new stream client interceptor that will decide whether to call
+// the interceptor based on the return argument from the Matcher.
+// TODO(bwplotka): Write unit test.
```
johanbrandhorst, line 72: Ditto

interceptors/selector/selector_example_test.go
```diff
@@ -6,14 +6,17 @@ package selector_test
 import (
 	"context"
 
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors"
+
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/selector"
```
johanbrandhorst, line 11: 😅

interceptors/selector/selector_test.go
```diff
@@ -7,37 +7,23 @@ import (
 	"context"
 	"testing"
 
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors"
```
johanbrandhorst, line 10: OK wait is this intentional?

interceptors/server_test.go
```diff
@@ -11,6 +11,8 @@ import (
 	"testing"
 	"time"
 
+	"google.golang.org/grpc/credentials/insecure"
```
johanbrandhorst, line 14: 🐽

providers/kit/examples_test.go
```diff
@@ -9,7 +9,7 @@ import (
 	"time"
 
 	"github.com/go-kit/log"
-	"github.com/grpc-ecosystem/go-grpc-middleware/providers/kit/v2"
+	"github.com/grpc-ecosystem/go-grpc-middleware/providers/kit"
```
johanbrandhorst, line 12: Import grouping. Also are we concerned that users will be confused by this not having v2 in its path?

providers/logr/examples_test.go
```diff
@@ -8,7 +8,7 @@ import (
 	"testing"
 	"time"
 
-	grpclogr "github.com/grpc-ecosystem/go-grpc-middleware/providers/logr/v2"
+	grpclogr "github.com/grpc-ecosystem/go-grpc-middleware/providers/logr"
```
johanbrandhorst, line 11: OK this is now becoming a pattern so; what is the rule for the third import group? I might suggest we just change it all to 2 import grouping instead of 3 to avoid this issue.

providers/logrus/examples_test.go
```diff
@@ -8,7 +8,7 @@ import (
 	"testing"
 	"time"
 
-	grpclogrus "github.com/grpc-ecosystem/go-grpc-middleware/providers/logrus/v2"
+	grpclogrus "github.com/grpc-ecosystem/go-grpc-middleware/providers/logrus"
```
johanbrandhorst, line 11: I'm gonna stop comment on imports.

providers/prometheus/doc.go
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) The go-grpc-middleware Authors.
+// Licensed under the Apache License 2.0.
+
+/*
+Package prometheus provides a standalone interceptor for metrics. It's next iteration of deprecated https://github.com/grpc-ecosystem/go-grpc-prometheus.
+See https://github.com/grpc-ecosystem/go-grpc-middleware/tree/main/examples for example.
```
johanbrandhorst, line 6: Our default branch is currently `master`, maybe we should change it to `main` first, and then merge v2 into it?

testing/testpb/interceptor_suite.go
```diff
@@ -17,6 +17,8 @@ import (
 	"sync"
 	"time"
 
+	"google.golang.org/grpc/credentials/insecure"
```
johanbrandhorst, line 20: OK this import is definitely in the wrong place

testing/testpb/pingservice_test.go
```diff
@@ -10,6 +10,11 @@ import (
 	"testing"
 	"time"
 
+	"google.golang.org/grpc/credentials/insecure"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
```
johanbrandhorst, line 16: 👀

johanbrandhorst (commented):
> Changing default branch to "v2" once v2.0.0 is out, should be fine.

I would prefer that we make a `v1` branch off `main` and merge `v2` into `main` instead. That's what we did for the gateway.

Do we have dependabot set up? With this module sprawl we'll want something to manage updates.

providers/logr/examples_test.go
bwplotka: Yea, simpler is better - will adjust.

providers/kit/examples_test.go
bwplotka: Yea, this is weird. I mentioned all in README. The alternative is to start sub modules with v2, which is also weird as `v2` at the end not in the middle like in the core module. So whatever we do, users will have to get used to multi-modules. Once you understand them it's much easier.

We can always release all modules and submodule with exactly the same version, that's an option.  (:

johanbrandhorst: Yeah, I agree I think we can keep this as is. It is a new module name and doesn't need a v2.

bwplotka:
I think I addressed all @johanbrandhorst

johanbrandhorst (approved):
Thanks a lot for this Bartek 🎉

PR #547, opened 2023-03-27
add support to trace on grpc_logrus.DefaultMessageProducer

### LINKED ISSUE #546: callback grpc_logrus.DefaultMessageProduce ignores TraceLevel
today if I use the option `grpc_logrus.WithLevels` and set one or more codes to be logged on `logrus.TraceLevel` I can't see the messages.

The root cause is `grpc_logrus.DefaultMessageProducer` that ignores TraceLevel

possible workaround: use `grpc_logrus.WithMessageProducer` to set a callback that can hand Trace

peczenyj:
today if I use the option `grpc_logrus.WithLevels` and set one or more codes to be logged on `logrus.TraceLevel` I can't see the messages.

The root cause is `grpc_logrus.DefaultMessageProducer` that ignores TraceLevel

possible workaround: use `grpc_logrus.WithMessageProducer` to set a callback that can hand Trace

this PR closes #546 546


johanbrandhorst (approved):

johanbrandhorst:
Thanks for this! Can you check if v2 also suffers from this bug?

peczenyj:
Hello

the branch v2 seems heavily different on this part, but I will try to check

PR #548, opened 2023-03-28
fix auto-generated docs

peczenyj:
somehow, after run `make all` there are some changes in all `doc.go` files. 

probably some auto fixes from some tool. since I'd like to submit another pull request, I dedided to extract this part that it is also common to #547 -- in case of the original pull request does not be accepted

enjoy


johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

PR #551, opened 2023-03-30
Add slog log provider

amenzhinsky:
Since `slog` is getting mainlined in `go1.21` it would be nice to start using it.


providers/slog/logger.go
```diff
@@ -0,0 +1,46 @@
...
+	case logging.INFO:
+		lvl = slog.LevelInfo
+	case logging.WARNING:
+		lvl = slog.LevelWarn
+	case logging.ERROR:
+		lvl = slog.LevelError
+	default:
+		panic(fmt.Sprintf("slog: unknown level %s", level))
+	}
+	l.Logger.Log(context.Background(), lvl, s)
```
amenzhinsky, line 45: BTW `Logger` interface could accept context too.

bwplotka:
Thanks! 

In the latest v2 we changed interface a lot (before v2.0.0) and we no longer maintain loggers. There is a logger interface (with context) and example uses https://github.com/grpc-ecosystem/go-grpc-middleware/tree/v2/interceptors/logging/examples meant to be copied. Feedback welcome, but this PR is no longer needed, sorry.

amenzhinsky:
@bwplotka no problem whatsoever, this is way better than having many adapters for each popular logger.

Issue #517, opened 2022-10-26
v2: interceptors/logging supports only strings as value


benammann:
During migration to `v2` we are expiring an issue that we cannot use nested or other types than string as custom log field value. We previously used `ctxzap` to add a `map[string]string` as log field value.

Current implementation using `v1`
```go
labels := map[string]string{
	"traceId": "insert-trace-id-here",
}
// wont compile because MarshalLogObject not implemented, should only serve as demonstration
ctxzap.AddFields(ctx, zap.Object("logging.googleapis.com/labels", labelValues))
```

We than expect the log to contain the following

````json
{
  "logging.googleapis.com/labels": {
    "traceId": "insert-trace-id-here"
  }
}
````

As zap does not use fields from ctxzap anymore we had to migrate to `logging.InjectFields` which leads to the following implementation

````go
type LogFieldValue interface {
	String() string
}

func InjectLogField(parentContext context.Context, key string, value LogFieldValue) context.Context {
	return logging.InjectFields(parentContext, logging.Fields{key, value.String()})
}
````

The problem with the new InjectFields method is that `logging.Fields` only allows `string` values so we cannot pass the nested data how we did before. The generated log now looks like this (using JSON Marshal)
````json
{
	"logging.googleapis.com/labels": "{\"traceId\":\"egym-cli-e99b4728-4d9a-4af9-8e75-56d0fe3e8a08\"}"
}
````

Is there any way we can archive to get the same output log as in v1 using the v2? If not, should we change the type of `logging.Fields` to accept more than just `strings` as value?

Thanks in advance

johanbrandhorst:
Hi, thanks for your issue report! Interested in hearing what Bartek thinks about this one 😄.

bwplotka:
Thanks for trying v2, and sorry for massive delay.

It seems indeed we have a slight limitation of Fields here. If your logger pushes JSON or complex objects our string requirement would flat all objects to quoted strings. 

We have to fix it, before v2.0 (huge PR in progress https://github.com/grpc-ecosystem/go-grpc-middleware/pull/543 for last mile end). Will check fix for this issue out once merged.

bwplotka:
I actually considering using slog Attributes and convert all to slog 🤔 https://pkg.go.dev/golang.org/x/exp/slog (since it was accepted in Go 1.21)

benammann:
@bwplotka awesome, really appreciate your work on this. Going to test this really soon.

PR #552, opened 2023-03-30
Simplified logging middleware; Fields are now "any" type; Moved logging providers to examples only.

bwplotka:
Fixes https://github.com/grpc-ecosystem/go-grpc-middleware/issues/517

* Simpler interface, only one method. This allows adapter to be ultra simple, so moved them to examples only (!).
* Fields are string, values are any.
* "More" slog compatibility.

I did not go for using slog natively as it's still experimental (see bigger discussion: https://gophers.slack.com/archives/C029RQSEE/p1679860885943619)


interceptors/logging/slog.go
```diff
@@ -0,0 +1,46 @@
...
+// Any system can map them to another numbering scheme if it wishes.
+//
+// First, we wanted the default level to be Info, Since Levels are ints, Info is
+// the default value for int, zero.
+//
+
+// Second, we wanted to make it easy to use levels to specify logger verbosity.
+// Since a larger level means a more severe event, a logger that accepts events
+// with smaller (or more negative) level means a more verbose logger. Logger
+// verbosity is thus the negation of event severity, and the default verbosity
```
yashrsharma44, line 27: Should using "inverse" would be a better option? Negation implies that our definition is tied up in the implementation(we are negating in this case), however using inverse means that lower the log level(enum) included, more verbose our logging is, which is simple to understand.

interceptors/logging/logging.go
```diff
@@ -119,70 +106,69 @@ func (f Fields) AppendUnique(add Fields) Fields {
...
+func InjectLogField(ctx context.Context, key string, val any) context.Context {
+	return InjectFields(ctx, Fields{key, val})
 }
 
-// Logger is unified interface that we used for all our interceptors. Official implementations are available under
-// provider/ directory as separate modules.
+// Logger requires Log method, similar to experimental slog, allowing logging interceptor to be interoperable. Official
+// adapters for popular loggers are in `provider/` directory (separate modules). It's totally ok to copy simple function
+// implementation over.
+// TODO(bwplotka): Once slog is official, we could use slog method directly. Currently level is copied over, so we don't
```
yashrsharma44, line 162: 👍

interceptors/logging/examples/kit/example_test.go
```diff
@@ -0,0 +1,70 @@
...
+	"google.golang.org/grpc"
+)
+
+// InterceptorLogger adapts go-kit logger to interceptor logger.
+// This code is simple enough to be copied and not imported.
+func InterceptorLogger(l log.Logger) logging.Logger {
+	return logging.LoggerFunc(func(_ context.Context, lvl logging.Level, msg string, fields ...any) {
+		largs := append([]any{"msg", msg}, fields...)
+		switch lvl {
+		case logging.LevelDebug:
```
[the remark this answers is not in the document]
bwplotka: What is ERROR? ERROR was replaced by LevelError and comon levels from slog there.

interceptors/logging/slog.go
bwplotka: I just copied all of this from slog, so there is less maintenance on our side (:  So if you don't like this you can talk to slog authors 🙈  (https://pkg.go.dev/golang.org/x/exp/slog#pkg-overview)... I feel we can use whatever, so I thought we could use what others most likely start doing more.

interceptors/logging/examples/kit/example_test.go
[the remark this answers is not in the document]
yashrsharma44: Okay, I meant we were using `logging.ERROR` before, but if slog uses that - [link](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/4402c19b3d262c7c77c3aa771ec3d322247b68f6/interceptors/logging/slog.go#L42), then let's use that convention.

yashrsharma44 (approved):

PR #555, opened 2023-04-04
Adjustments to README and consistency of callback options.

bwplotka:


bwplotka:
It's small one so progressing with self-review.

Issue #559, opened 2023-04-11
Is this a bug? zap logging 


wrapupc:
`logger := zap.NewExample()    
return grpc.ChainUnaryInterceptor(
		logging.UnaryServerInterceptor(InterceptorZapLogger(logger)),
	)`

I added zap into grpc interceptor, and here is what i got after  a request

`
{"level":"info","msg":"started call","protocol":"grpc","protocol":"grpc","protocol":"grpc","protocol":"grpc","protocol":"grpc","protocol":"grpc","protocol":"grpc","protocol":"grpc"}
{"level":"info","msg":"finished call","protocol":"grpc","protocol":"grpc","protocol":"grpc","protocol":"grpc","protocol":"grpc","protocol":"grpc","protocol":"grpc","protocol":"grpc","protocol":"grpc","p
rotocol":"grpc","protocol":"grpc","protocol":"grpc","protocol":"grpc","protocol":"grpc","protocol":"grpc","protocol":"grpc","protocol":"grpc"}
`

bwplotka:
Hi, can you check latest tag (https://github.com/grpc-ecosystem/go-grpc-middleware/releases/tag/v2.0.0-rc.5)? We recommend now to create your own "logging.Logger" implementation, see https://github.com/grpc-ecosystem/go-grpc-middleware/tree/main/interceptors/logging/examples/zap

Let me know if you can still repro the problem with this new approach.

bwplotka:
I guess you actually found it in new version and fixed, nice!

https://github.com/grpc-ecosystem/go-grpc-middleware/pull/565#pullrequestreview-1383162951

marcoshuck:
> Hi, can you check latest tag (https://github.com/grpc-ecosystem/go-grpc-middleware/releases/tag/v2.0.0-rc.5)? We recommend now to create your own "logging.Logger" implementation, see https://github.com/grpc-ecosystem/go-grpc-middleware/tree/main/interceptors/logging/examples/zap
> 
> Let me know if you can still repro the problem with this new approach.

What's the motivation behind this? I've found myself copy-pasting this code multiple times in different projects.

PR #565, opened 2023-04-12
fix: iter would no longer be inited every time in the loop

wrapupc:
fix: iter would no longer be inited every time in the loop

## Verification

I've used this code in my project, and the log looks good now

### DIFF
--- a/interceptors/logging/examples/zap/example_test.go
+++ b/interceptors/logging/examples/zap/example_test.go
@@ -17,10 +17,10 @@ import (
 func InterceptorLogger(l *zap.Logger) logging.Logger {
 	return logging.LoggerFunc(func(ctx context.Context, lvl logging.Level, msg string, fields ...any) {
 		f := make([]zap.Field, 0, len(fields)/2)
+		iter := logging.Fields(fields).Iterator()
 		for i := 0; i < len(fields); i += 2 {
-			i := logging.Fields(fields).Iterator()
-			if i.Next() {
-				k, v := i.At()
+			if iter.Next() {
+				k, v := iter.At()
 				f = append(f, zap.Any(k, v))
 			}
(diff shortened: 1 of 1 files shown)


interceptors/logging/examples/zap/example_test.go
```diff
@@ -17,10 +17,10 @@ import (
 func InterceptorLogger(l *zap.Logger) logging.Logger {
 	return logging.LoggerFunc(func(ctx context.Context, lvl logging.Level, msg string, fields ...any) {
 		f := make([]zap.Field, 0, len(fields)/2)
+		iter := logging.Fields(fields).Iterator()
 		for i := 0; i < len(fields); i += 2 {
```
bwplotka, line 21: I think it can be simplified further, similar to https://github.com/grpc-ecosystem/go-grpc-middleware/blob/main/interceptors/logging/examples/logrus/example_test.go#L21

I messed this up 🙃  Can we double check other examples too

bwplotka (approved):
Nice, thanks <3

One nit, otherwise LGTM - also CLA is required. Thanks!!

bwplotka:
Fixes https://github.com/grpc-ecosystem/go-grpc-middleware/issues/559

bwplotka:
Also we need tests, but we can do them in later PRs (or here) - added issue to track this: https://github.com/grpc-ecosystem/go-grpc-middleware/issues/567

interceptors/logging/examples/zap/example_test.go
chancez: It actually needs to be `for iter.Next()`. This issue seems to be present on all the examples.

chancez:
This should be fixed via #588 so I think this can be closed.

PR #569, opened 2023-04-16
fix: correct example to avoid zerolog fields combined

levanlongktmt:
I follow the current code in the example of using `zerolog` and `fields` grown after each request. It combines all `fields` to the context of `l` after each request and makes the log messages become huge. 
It's the same with this issue on `zap` https://github.com/grpc-ecosystem/go-grpc-middleware/issues/559

image

I fixed this by creating a new instance of `zerolog.Logger` instead of using `l` and it works like a charm. 

image

* [x] I added CHANGELOG entry for this change.

## Changes

* [x] Correct example of `interceptors/logging/examples/zerolog`

## Verification

I tested and it works on my project, logs work fine and didn't get combined `fields`


johanbrandhorst (approved):

johanbrandhorst:
Thanks for the fix!

PR #571, opened 2023-04-21
fix prometheus interceptors not converting context errors to gRPC codes

vtermanis:
First of all - thank you for a great library!

---

* [ ] I added CHANGELOG entry for this change. (**I don't see a changelog file - what am I missing?**)

## Changes
fix prometheus interceptors not converting context errors to gRPC codes
- Previously `context.(Canceled|DeadlineExceeded)` would result in an "Unknown" grpc_code.

We use the interceptors to not only track failures but also identify when we've accidentally passed back non-grpc errors. With this fix the context errors are correctly mapped to their gRPS status counterparts. This behaves the same as what happens in the [golang server code](https://github.com/grpc/grpc-go/blob/fe72db9589696a625af72117a600ba191227b7c5/server.go#L1339-L1345). Something similar but old existed in your deprecated repo (https://github.com/grpc-ecosystem/go-grpc-prometheus/pull/84) but it was never released nor did it make it across, it would appear.

## Verification

Added unit test which triggers `context.Canceled` to verify it's no longer recorded as `grpc_code="Unknown"`

### DIFF
--- a/providers/prometheus/options.go
+++ b/providers/prometheus/options.go
@@ -9,11 +9,15 @@ import (
 	"google.golang.org/grpc/status"
 )
 
-// FromError returns a grpc status if error code is a valid grpc status.
-func FromError(err error) (s *status.Status, ok bool) {
-	return status.FromError(err)
-
-	// TODO: @yashrsharma44 - discuss if we require more error handling from the previous package
+// FromError returns a grpc status. If the error code is neither a valid grpc status nor a context error, codes.Unknown
+// will be set.
+func FromError(err error) *status.Status {
+	s, ok := status.FromError(err)
+	// Mirror what the grpc server itself does, i.e. also convert context errors to status
+	if !ok {
+		s = status.FromContextError(err)
+	}
+	return s
 }
 
 // A CounterOption lets you add options to Counter metrics using With* funcs.
--- a/providers/prometheus/reporter.go
+++ b/providers/prometheus/reporter.go
@@ -22,7 +22,7 @@ type reporter struct {
 
 func (r *reporter) PostCall(err error, rpcDuration time.Duration) {
 	// get status code from error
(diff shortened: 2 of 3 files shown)


johanbrandhorst (changes requested):
Hi, could you make this change against v2? v1 is kinda maintenance mode.

vtermanis:
@johanbrandhorst - is this what you meant? (I didn't realise v2 branch was current given `main` also had activity on it)

vtermanis:
The unit & lint test checks don't run for some reason. Any chance one of the maintainers could take a look why that is? (I don't believe I have access to see.)

vtermanis:
@johanbrandhorst  - apologies for the ping but it would be great to for this to be considered.

The unit & lint test checks don't run for some reason (they do pass). Any chance you or one of the maintainers could take a look why that is? (I don't believe I have access to see/re-run the tests)

johanbrandhorst:
> @johanbrandhorst - apologies for the ping but it would be great to for this to be considered.
> 
> The unit & lint test checks don't run for some reason (they do pass). Any chance you or one of the maintainers could take a look why that is? (I don't believe I have access to see/re-run the tests)

Sorry to have missed this - we merged v2 into main so this should be targeted at main now. Please update again 😅.

vtermanis:
note: Do **not** merge - the added failing test still shows the suite as passing (which obv. is wrong). I'll try to figure out what is wrong.

vtermanis:
Ok, now it passes:
- Corrected labels to use `bidi_stream` (instead of `server_stream`)
- Fixed typo of `Canceled` (was `Cancelled`)

Ready for review

conbon (approved):

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

Issue #567, opened 2023-04-13
Add tests to interceptors/logging/examples


bwplotka:
We need unit tests for https://github.com/grpc-ecosystem/go-grpc-middleware/tree/main/interceptors/logging/examples, so we can avoid bugs like https://github.com/grpc-ecosystem/go-grpc-middleware/issues/559

rohanraj7316:
sorry, missed out here [565](https://github.com/grpc-ecosystem/go-grpc-middleware/pull/565#pullrequestreview-1383162951) is already in PR. do we need to replicate that here? @bwplotka

bwplotka:
I mean PR is not moving, so if you can actually fix it / add test that would be helpful. We need for all, but even one would be helpful 💪🏽

rohanraj7316:
just a bit of thought here. should I rename `example_test.go` to `example.go` and then add test case in `example_test.go` just a thought to improve naming convention or if you have anything better pls let me know @bwplotka

bwplotka:
Sgtm!

PR #577, opened 2023-05-08
fix zap logger example and add tests

theruziev:

## Changes

- Fix zap logging example 
- Add a test for zap example

Before: 
Screenshot 2023-05-08 at 16 16 49

After:
Screenshot 2023-05-08 at 16 19 28

## Verification

I added test for testing/verification

### DIFF
--- a/interceptors/logging/examples/kit/example_test.go
+++ b/interceptors/logging/examples/kit/example_test.go
@@ -4,12 +4,20 @@
 package kit_test
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"runtime"
+	"strings"
+	"testing"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/logging"
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/testing/testpb"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
 	"google.golang.org/grpc"
 )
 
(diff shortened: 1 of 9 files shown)


theruziev:
I don't understand why linter failed. I added a copyright header

theruziev:
I found discussion (#567) about adding tests for logging examples and add it. 
@bwplotka cc

interceptors/logging/examples/logrus/example.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+// InterceptorLogger adapts logrus logger to interceptor logger.
+// This code is simple enough to be copied and not imported.
+func InterceptorLogger(l logrus.FieldLogger) logging.Logger {
+	return logging.LoggerFunc(func(_ context.Context, lvl logging.Level, msg string, fields ...any) {
+		f := make(map[string]any, len(fields)/2)
+		i := logging.Fields(fields).Iterator()
+		for i.Next() {
```
theruziev, line 20: https://github.com/grpc-ecosystem/go-grpc-middleware/blob/main/interceptors/logging/examples/logrus/example_test.go#L21 - before

bwplotka:
Thanks for this and sorry for lag. Example was fixed, but I would love those tests! Do you mind rebasing?

theruziev:
yep

theruziev:
rebased, and returned examples function, maybe it's helpful to see how used it.

theruziev:
ping @bwplotka @johanbrandhorst

interceptors/logging/examples/kit/example_test.go
```diff
@@ -58,13 +45,59 @@ func ExampleInterceptorLogger() {
...
 		grpc.WithChainStreamInterceptor(
-			logging.StreamClientInterceptor(InterceptorLogger(logger), opts...),
+			logging.StreamClientInterceptor(examplekit.InterceptorLogger(logger), opts...),
 			// Add any other interceptor you want.
 		),
 	)
 	// Output:
 }
+
+type kitExampleTestSuite struct {
```
bwplotka, line 59: Let's avoid "suite" patterns, especially for single method test. Why not single unit test?

bwplotka (changes requested):
Thanks for this!

Tests are great (just I would move out of suite, but not a blocker). 

However we kept InterceptorLogger adapters in test *exactly* so no one depends / import those. It's only for inspiration and copy. There is even comment suggesting that `// This code is simple enough to be copied and not imported.`. We have to enforce this otherwise people will start to depend on this module which is not meant to be specially versioned and compatibile (!).

Can we move those back?

interceptors/logging/examples/kit/example_test.go
theruziev: I used the test suite because that already has a server and client I can easily use for a test. I try to find another way to do that

theruziev:
I returned interceptors to test files

interceptors/logging/examples/kit/example_test.go
theruziev: if you have any idea how to that this without test suite I will try to implement that.

GreyYar: > Let's avoid "suite" patterns, especially for single method test. Why not single unit test?

Hello, pardon my curiosity but I wonder why should we avoid suite pattern since it gave us a server and client inside?
It gave us expandable way to add more tests in the future.

johanbrandhorst: Test suite types in Go like this are uncommon and unnecessary. If you need a server in a test, you can just create one. If you want to isolate the logic, create a function. Test suites take OOP beyond what is usually necessary in Go. Just write test functions, skip the suite.

johanbrandhorst (changes requested):
Rewrite the "test suite" to one or more separate tests.

theruziev:
i work on it

PR #581, opened 2023-05-11
Changed for for mapping fields, different var for logger in zap example

MichalFikejs:
* [x] I added CHANGELOG entry for this change.

## Changes

* Mapping Fields works correctly now
* Logger instance used for logging is now being used only once, previous code has been rewriting instance, and then duplicating fields.

## Verification
I tested it and using it in my project :)


johanbrandhorst (approved):
LGTM

johanbrandhorst:
Thanks for your contribution!

PR #584, opened 2023-05-13
providers/prometheus: Add WithHistogramOpts for native histograms

metalmatze:

Where exactly do I need to add the CHANGELOG?

## Changes

We need to be able to set the NativeHistogramBucketFactor for example to enable native histograms. Therefore, the `WithHistogramBuckets` Option isn't flexible enough anymore. I've added the `WithHistogramOpts` option.

## Verification

grpc-nativehistograms


johanbrandhorst (approved):
Seems reasonable to me!

johanbrandhorst:
Thanks for your contribution!

PR #586, opened 2023-05-26
Minor code cleanups

ash2k:

## Changes

- Use `strings.Cut()` as it's more readable and doesn't allocate a slice.
- Replace `Errorf()` with `Error()` where no formatting is being done.

## Verification

Unit tests.

### DIFF
--- a/examples/server/main.go
+++ b/examples/server/main.go
@@ -111,7 +111,7 @@ func main() {
 		}
 		// TODO: This is example only, perform proper Oauth/OIDC verification!
 		if token != "yolo" {
-			return nil, status.Errorf(codes.Unauthenticated, "invalid auth token")
+			return nil, status.Error(codes.Unauthenticated, "invalid auth token")
 		}
 		// NOTE: You can also pass the token in the context for further interceptors or gRPC service code.
 		return ctx, nil
--- a/interceptors/auth/auth_test.go
+++ b/interceptors/auth/auth_test.go
@@ -39,7 +39,7 @@ func buildDummyAuthFunction(expectedScheme string, expectedToken string) func(ct
 			return nil, err
 		}
 		if token != expectedToken {
-			return nil, status.Errorf(codes.PermissionDenied, "buildDummyAuthFunction bad token")
+			return nil, status.Error(codes.PermissionDenied, "buildDummyAuthFunction bad token")
 		}
 		return context.WithValue(ctx, authedMarker, "marker_exists"), nil
 	}
--- a/interceptors/auth/metadata.go
+++ b/interceptors/auth/metadata.go
@@ -24,14 +24,14 @@ var (
 func AuthFromMD(ctx context.Context, expectedScheme string) (string, error) {
 	val := metadata.ExtractIncoming(ctx).Get(headerAuthorize)
 	if val == "" {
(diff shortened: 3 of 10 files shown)


interceptors/ratelimit/ratelimit.go
```diff
@@ -22,7 +22,7 @@ type Limiter interface {
 func UnaryServerInterceptor(limiter Limiter) grpc.UnaryServerInterceptor {
 	return func(ctx context.Context, req any, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (any, error) {
 		if err := limiter.Limit(ctx); err != nil {
-			return nil, status.Errorf(codes.ResourceExhausted, "%s is rejected by grpc_ratelimit middleware, please retry later. %s", info.FullMethod, err)
+			return nil, status.Errorf(codes.ResourceExhausted, "%s is rejected by grpc_ratelimit middleware, please retry later. %v", info.FullMethod, err)
```
johanbrandhorst, line 25: I don't think this is an improvement - the value in using specific verbs (like `%s`) over generic (like `%v`) is that go vet can warn when you're using the specific verbs incorrectly. Please revert this change.

johanbrandhorst (commented):
Want to fix the lint warning at the same time :)?

interceptors/ratelimit/ratelimit.go
ash2k: Hm, I'm not sure I understand your concern. Isn't it idiomatic to use `%v` for the `error` type formatting? 99% of code does it that way, including in this codebase.

ash2k: At the same time, there is code that does the opposite of what you are suggesting - uses `%v` for e.g. `float32`:

https://github.com/ash2k/go-grpc-middleware/blob/0df73fc576befff7acf6f397634a64c54aa4ad11/interceptors/logging/options.go#L81-L94

johanbrandhorst: I don't think that's an argument against what I'm saying, it'd be great to change those instances to not use %v either. I would only use %v where I truly don't know the type.

ash2k: I'm not sure if you've seen my previous comment? `%v` is always used for errors (I mean, check stdlib?), it even handles them in a [special way](https://pkg.go.dev/fmt):

```
4. If an operand implements the error interface, the Error method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any). 
```

johanbrandhorst: That's not a special case for the %v verb, it goes for any verb. All the more reason to use %s for errors, so we get the benefit of `go vet` checking our interpolations for us.

ash2k: That's interesting! I didn't know this special case is not just for `error`. Ok, I changed it back to `%s`.

johanbrandhorst (changes requested):
Thanks for the updates! Sorry I just realized this is targeted at the v2 branch, but we've merged v2 into main so please target this at main.

ash2k:
Looks like the CLA check is not working correctly when branch is pushed to with a new base commit...

johanbrandhorst (approved):

johanbrandhorst:
Thank you!

PR #588, opened 2023-06-06
Fix logrus interceptor

chancez:

## Changes

- Fixes the example logrus interceptor so that it actually includes all log fields
- Updates zap logger to use the logging fields iterator correctly, removing redundant logic. Subsumes https://github.com/grpc-ecosystem/go-grpc-middleware/pull/565

## Verification

I've tested logrus in my own project, but I haven't fully tested the zap changes, though it should be equivalent.


johanbrandhorst (approved):
Nice, thanks

PR #608, opened 2023-08-02
Use default go errors package instead of github.com/pkg/errors

rifkyazizf:
## Changes
- change code that use github.com/pkg/errors
- remove `errors=github.com/pkg/errors` in makefile 

## Verification
Unit tests.


johanbrandhorst (approved):
Thanks, this is great

PR #612, opened 2023-08-08
fix bug : should drain channel of timer after stop

ikenchina:
Ensure the channel is empty after a call to Stop.

## Changes

Drain channel.

## Verification

Document of go Timer notices that : `To ensure the channel is empty after a call to Stop, check the
// return value and drain the channel.`

### DIFF
--- a/interceptors/retry/retry.go
+++ b/interceptors/retry/retry.go
@@ -260,7 +260,9 @@ func waitRetryBackoff(attempt uint, parentCtx context.Context, callOpts *options
 		timer := time.NewTimer(waitTime)
 		select {
 		case <-parentCtx.Done():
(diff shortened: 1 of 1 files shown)


ikenchina:
update

johanbrandhorst:
Have you been able to sign the CLA?

ikenchina:
Yes

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

ash2k:
The timer is discarded after this use, it doesn't matter if the channel has a value or not. No need to drain it :)

win5do:
There is no bug, this commit msg and changes are puzzling.

PR #628, opened 2023-08-29
feat: Support extracting fields from CallMeta

fsaintjacques:
## Changes

Extend `logging.WithFieldsFromContext` via a new `logging.WithFieldsFromContextAndCallMeta` in a backward compatible fashion such that it supports extracting fields not only from the Context but also from the CallMeta.

## Verification

I fixed the original unit test and after the change it still passes.


fsaintjacques:
CLA is signed.

johanbrandhorst (approved):
This makes sense to me, thanks!

Issue #604, opened 2023-07-13
I want to ignore some defult logged fields 


mehmetakiftatar:
I use github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors for add logging , but it came with different unnesesery fields like `grpc.method_type` i dont need thoose. So i forked repo and customize it. But i think we need option to ignore given fields with key. 

**What happened**:
... message="started call" protocol=grpc grpc.component=server grpc.service=MockService grpc.method=GetUser grpc.method_type=unary peer.address=127.0.0.1:35722 grpc.start_time=2023-07-13T09:32:34Z grpc.time_ms=0.019

**What you expected to happen**:
... message="started call" grpc.method=GetUser 

If accept the issue i can open pr for change the behavior.

bwplotka:
Makes sense to me. Customization here would make sense. Perhaps option to specify fields to add would work (with default specifying all)

PR #631, opened 2023-09-05
Add logging option to disable fields in log entry

coleenquadros:
Ref issue - https://github.com/grpc-ecosystem/go-grpc-middleware/issues/604
Added a logging option to allow user to select fields from the common set of grpc log fields to be ignored/disabled to the log entry -


interceptors/logging/options.go
```diff
@@ -204,3 +206,21 @@ func WithTimestampFormat(format string) Option {
...
+//
+// If not provided default fields will be used.
+type AddGrpcLogFields func() []string
+
+// WithAddGrpcLogFields customizes the function for adding gRPC fields to the log entry.
+func WithGrpcLogFields(f AddGrpcLogFields) Option {
+	return func(o *options) {
+		o.grpcLogFields = f
+	}
+}
```
johanbrandhorst, line 226: This seems like a complicated way to ignore fields from the default fields. I think just a way to pass the existing field names might be easier?

```go
func WithDisableLoggingField(disabledLogFields ...string) Option {
	return func(o *options) {
		o.disabledLogFields = disabledLogFields
	}
```

Then if can be used like so:

```go
[]grpc.DialOption{
		grpc.WithUnaryInterceptor(logging.UnaryClientInterceptor(s.logger, logging.WithGrpcLogFields(logging.MethodFieldKey, logging.MethodTypeFieldKey))),
	}
```

What do you think?

coleenquadros: Yes, that would be better.  Ive updated the commit. :)

johanbrandhorst (changes requested):
Sorry, I think I must've confused you. I think this new setting should be used for _disabling_ logging fields, not enabling. So if I don't specify the option, it should include all default logging fields. If I specify the option, it should remove any fields I specified from the logger. If I specify the option with a field that's not used, then it will do nothing. That's why my suggestion said `WithDisableLoggingFields`. I think if we use this code as suggested it will make the behaviour confusing for users.

coleenquadros:
@johanbrandhorst. Oh alright. I assumed from the [comment](https://github.com/grpc-ecosystem/go-grpc-middleware/issues/604#issuecomment-1685360628) in the issue was to provide an option for the fields to be added. I will change it to disable

interceptors/logging/interceptors.go
```diff
@@ -141,8 +141,15 @@ func reportable(logger Logger, opts *options) interceptors.CommonReportableFunc
...
 
-		// Field dups from context override the common fields.
-		fields := newCommonFields(kind, c).WithUnique(ExtractFields(ctx))
+		fields := Fields{}
+		if opts.disableGrpcLogFields != nil {
+			fields = disableCommonLoggingFields(kind, c, opts.disableGrpcLogFields)
+		} else {
+			// Field dups from context override the common fields.
+			fields = newCommonFields(kind, c)
+		}
```
johanbrandhorst, line 150: You can rewrite this to avoid the else
```suggestion
		// Field duplicates from the context may override the common fields.
		fields := newCommonFields(kind, c)
		if opts.disableGrpcLogFields != nil {
			fields = disableCommonLoggingFields(kind, c, opts.disableGrpcLogFields)
		}
```

interceptors/logging/logging.go
```diff
@@ -38,6 +38,30 @@ func newCommonFields(kind string, c interceptors.CallMeta) Fields {
...
+	i := commonFields.Iterator()
+	for i.Next() {
+		k, v := i.At()
+		existing[k] = v
+	}
+
+	for _, key := range disableFields {
+		if _, ok := existing[key]; ok {
+			delete(existing, key)
+		}
```
johanbrandhorst, line 55: You can just call delete without checking if the key exists

interceptors/logging/logging.go
```diff
@@ -38,6 +38,30 @@ func newCommonFields(kind string, c interceptors.CallMeta) Fields {
 	}
 }
 
+// disableCommonLoggingFields returns copy of common fields with disabled fields removed.
+func disableCommonLoggingFields(kind string, c interceptors.CallMeta, disableFields []string) Fields {
+	commonFields := newCommonFields(kind, c)
+
+	existing := map[any]any{}
```
johanbrandhorst, line 45: Can't this be of the `Fields` type? Then we don't have to do the final loop at the end.

interceptors/logging/options.go
```diff
@@ -204,3 +206,10 @@ func WithTimestampFormat(format string) Option {
 		o.timestampFormat = format
 	}
 }
+
+// WithDisableLoggingFields disables logging of gRPC fields provided.
```
johanbrandhorst, line 210: Add some reference to where the default logging fields are defined, and also add an example use case here please.

johanbrandhorst (changes requested):
There are lint errors, please see the failing action and fix them.

interceptors/logging/logging.go
coleenquadros: I took the reference from here -> https://github.com/grpc-ecosystem/go-grpc-middleware/blob/4679fb12b6915f8f7a682a525073fe3810d5c64e/interceptors/logging/logging.go#L81

Fields is of `type Fields []any` and delete does not work on that type

johanbrandhorst: I see. I think there are still a few problems with this solution:
1. The current solution will reorder elements in the fields (because of the map iteration), which could be confusing and annoying to users.
2. It might be easier to just populate `newFields` directly while iterating over `disableFields`.

I think the better solution here is actually to implement a new function on `Fields`: `Delete(k string)`. `Fields` is essentially pretending to be a map, so deleting by a key makes sense. The implementation of the `Delete` method would mutate the receiver directly by removing the element and the value immediately following it. Then this function can just loop over the disabledFields and call Delete.

What do you think?

johanbrandhorst (changes requested):

interceptors/logging/logging.go
```diff
@@ -38,6 +38,22 @@ func newCommonFields(kind string, c interceptors.CallMeta) Fields {
...
 }
 
+// disableCommonLoggingFields returns copy of newCommonFields with disabled fields removed from the following
+// default list.
+//
+//	-SystemTag[0],
+//	-ComponentFieldKey,
+//	-ServiceFieldKey,
+//	-MethodFieldKey,
+//	-MethodTypeFieldKey,
```
johanbrandhorst, line 48: Go comment lists have [a standard way](https://tip.golang.org/doc/comment) of being formatted. I think this should be
```suggestion
// disableCommonLoggingFields returns copy of newCommonFields with disabled fields removed from the following
// default list. The following are the default logging fields:
//   - SystemTag[0]
//   - ComponentFieldKey
//   - ServiceFieldKey
//   - MethodFieldKey
//   - MethodTypeFieldKey
```

johanbrandhorst (changes requested):
Thanks, almost there! Could you add some tests for the new Delete method, to ensure it acts correctly? Thanks!

johanbrandhorst (approved):
LGTM, thanks!

PR #651, opened 2023-10-03
Update README.md - re metric names inherited from https://github.com/grpc-ecosystem/go-grpc-prometheus

bwplotka:
Thanks @brancz for reminding us to mention this in docs!


brancz (approved):
Thank you very much for double checking!

PR #660, opened 2023-10-13
add idempotency key

idelvall:
Adds an idempotency key to the headers sent by the requests and its retries so the server can use this header to avoid processing multiple times a non-idempotent implementation.

> Idempotency keys are unique tokens that you submit as a request header, that guarantee that only one resource will be created regardless of how many times a request is sent to us.

### DIFF
--- a/interceptors/retry/retry.go
+++ b/interceptors/retry/retry.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/grpc-ecosystem/go-grpc-middleware/v2/metadata"
 	"golang.org/x/net/trace"
 	"google.golang.org/grpc"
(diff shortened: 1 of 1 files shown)


johanbrandhorst (changes requested):
This seems like a reasonable idea, but couldn't it just be another interceptor? Why does it need to be part of the retry logic? Also please add some tests.

idelvall:
> This seems like a reasonable idea, but couldn't it just be another interceptor? Why does it need to be part of the retry logic? Also please add some tests.

Hi @johanbrandhorst, yes I considered that option, but my reasoning was that given that non-idempotent RPCs should not be retried, this is something that falls into the library scope.

Happy to add those tests if you think the change is wroth it.

johanbrandhorst:
Ah, you mean that the use of the `retry` interceptor _implies_ that the request must be idempotent? I suppose I follow that logic.

johanbrandhorst:
Yes, please do add the tests and fix the build, thanks

PR #670, opened 2023-10-31
adds fields from durationFieldFunc to request/response log entries

vroldanbet:
👋🏻 Hai good people of the `go-grpc-middleware`

## Changes

This PR adds the fields resulting from invoking `durationFieldFunc` to request and response log messages 

## Why

the request and response log messages include a duration string but do not use the `durationFieldFunc` like other parts of the request lifecycle.

The `durationFieldFunc` function is useful when conducting log analysis with log aggregator tools because it normalizes the value to a number that can be filtered by or compared to. For example, a GCP Log Explorer [does not support converting log field types like duration to an integer](https://stackoverflow.com/a/63746056), so you cannot run queries like "give me all logs that took more than 50ms" because it would be interpreted as a string comparison. This makes it very tedious to audit application logs that log API calls using this gRPC middleware.

The perhaps controversial part of this change is that request/response logs have a different naming convention for the duration. I can't find a use for a different field name - `grpc.time` and `grpc.time_ms` would seem expressive enough to support the 4 logged entries, but I'm likely missing something here. I would suggest renaming those fields as well, but I understand that may be a "breaking change", if we consider the log fields a contract.

## Verification

The unit tests failed immediately and I then updated them.

### DIFF
--- a/interceptors/logging/interceptors.go
+++ b/interceptors/logging/interceptors.go
@@ -71,6 +71,7 @@ func (c *reporter) PostMsgSend(payload any, err error, duration time.Duration) {
 		}
 
 		fields = fields.AppendUnique(Fields{"grpc.send.duration", duration.String(), "grpc.request.content", p})
+		fields = fields.AppendUnique(c.opts.durationFieldFunc(duration))
 		c.logger.Log(c.ctx, logLvl, "request sent", fields...)
 	} else {
 		p, ok := payload.(proto.Message)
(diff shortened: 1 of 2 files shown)


johanbrandhorst (approved):
This seems reasonable to me, though I wonder if it might confuse users to have both `grpc.send.duration` and this new user configurable field? Ideally I'd see some way for the user to explicitly enable this, but it seems the duration field func has a default, so we can't just check if it's set or not.

johanbrandhorst:
I agree that we should not rename the existing default fields as it may break users.

PR #682, opened 2023-11-25
Middleware for determining the real ip of the client

MadsRC:
## Changes

This PR adds a middleware which will determine the real ip / original IP of the client based on header values and inject this into the context of the request. A function is also introduced to allow retrieval of this value from the context at a later stage.

When a server is behind a load balancer it is often useful to attempt to determine the IP of the original client so as to not only log the IP of the load balancer. Solutions such as NGINX, AWS ELB and Google Cloud Load Balancers all provide this by appending the IP of the client to an `x-forwarded-for` header. Cloudflare provides this by setting the `True-Client-IP` header to the value of the client.

Only gRPC servers being served over TCP/IP networks are supported.

Implementation of this functionality carries with it the risk of trusting unverified user input and as such allowing the spoofing of source IP's or Denial of Service attacks. Because of this I believe this project should provide a verified and trusted implementation. Such an implementation is included in this PR.

I opted to use a table-driven approach to testing this - I'm unsure if this is acceptable for this repo. If not, please let me know and I'll refactor it.

## Verification

Unit testing with 100% coverage has been implemented. Additionally, I believe I've covered the majority of the edge-cases in testing.

From a security standpoint, the tests verifies that headers are only referenced if the request is coming from a trusted peer. In the event that a trusted peer sends a malformed header, the header is rejected (which is also covered by the tests).

### DIFF
--- a/interceptors/realip/examples_test.go
+++ b/interceptors/realip/examples_test.go
@@ -1,0 +1,43 @@
+// Copyright (c) The go-grpc-middleware Authors.
+// Licensed under the Apache License 2.0.
+
+package realip_test
+
+import (
+	"net"
+
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/realip"
+	"google.golang.org/grpc"
+)
+
+// Simple example of a unary server initialization code.
+func ExampleUnaryServerInterceptor() {
+	// Define list of trusted peers from which we accept forwarded-for and
+	// real-ip headers.
+	trustedPeers := []net.IPNet{
+		{IP: net.IPv4(127, 0, 0, 1), Mask: net.IPv4Mask(255, 255, 255, 255)},
+	}
+	// Define headers to look for in the incoming request.
+	headers := []string{"x-forwarded-for", "x-real-ip"}
+	_ = grpc.NewServer(
+		grpc.ChainUnaryInterceptor(
+			realip.UnaryServerInterceptor(trustedPeers, headers),
+		),
+	)
+}
+
+// Simple example of a streaming server initialization code.
+func ExampleStreamServerInterceptor() {
+	// Define list of trusted peers from which we accept forwarded-for and
+	// real-ip headers.
+	trustedPeers := []net.IPNet{
+		{IP: net.IPv4(127, 0, 0, 1), Mask: net.IPv4Mask(255, 255, 255, 255)},
+	}
+	// Define headers to look for in the incoming request.
+	headers := []string{"x-forwarded-for", "x-real-ip"}
+	_ = grpc.NewServer(
+		grpc.ChainStreamInterceptor(
+			realip.StreamServerInterceptor(trustedPeers, headers),
+		),
+	)
+}
--- a/interceptors/realip/examples_test.go
+++ b/interceptors/realip/examples_test.go
@@ -4,7 +4,7 @@
 package realip_test
 
 import (
-	"net"
+	"net/netip"
 
 	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/realip"
 	"google.golang.org/grpc"
@@ -14,11 +14,11 @@ import (
 func ExampleUnaryServerInterceptor() {
 	// Define list of trusted peers from which we accept forwarded-for and
 	// real-ip headers.
-	trustedPeers := []net.IPNet{
-		{IP: net.IPv4(127, 0, 0, 1), Mask: net.IPv4Mask(255, 255, 255, 255)},
+	trustedPeers := []netip.Prefix{
+		netip.MustParsePrefix("127.0.0.1/32"),
 	}
 	// Define headers to look for in the incoming request.
-	headers := []string{"x-forwarded-for", "x-real-ip"}
+	headers := []string{realip.X_FORWARDED_FOR, realip.X_REAL_IP}
 	_ = grpc.NewServer(
 		grpc.ChainUnaryInterceptor(
 			realip.UnaryServerInterceptor(trustedPeers, headers),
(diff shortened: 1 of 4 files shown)


interceptors/realip/examples_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+// Simple example of a unary server initialization code.
+func ExampleUnaryServerInterceptor() {
+	// Define list of trusted peers from which we accept forwarded-for and
+	// real-ip headers.
+	trustedPeers := []net.IPNet{
+		{IP: net.IPv4(127, 0, 0, 1), Mask: net.IPv4Mask(255, 255, 255, 255)},
+	}
+	// Define headers to look for in the incoming request.
+	headers := []string{"x-forwarded-for", "x-real-ip"}
```
johanbrandhorst, line 21: What do you think about defining these header name as constants in the package, to make it easier?

interceptors/realip/realip.go
```diff
@@ -0,0 +1,116 @@
...
+type serverStream struct {
+	grpc.ServerStream
+	ctx context.Context
+}
+
+func (s *serverStream) Context() context.Context {
+	return s.ctx
+}
+
+// UnaryServerInterceptor returns a new unary server interceptors that performs request rate limiting.
```
johanbrandhorst, line 93: I think this doc string will need updating 😁

interceptors/realip/realip.go
```diff
@@ -0,0 +1,116 @@
...
+	return func(ctx context.Context, req any, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (any, error) {
+		ip := getRemoteIP(ctx, trustedPeers, headers)
+		if ip != nil {
+			ctx = context.WithValue(ctx, realipKey{}, ip)
+		}
+		return handler(ctx, req)
+	}
+}
+
+// StreamServerInterceptor returns a new stream server interceptor that performs rate limiting on the request.
```
johanbrandhorst, line 104: Ditto

interceptors/realip/realip.go
```diff
@@ -0,0 +1,116 @@
...
+	"strings"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/peer"
+)
+
+type realipKey struct{}
+
+func FromContext(ctx context.Context) (net.IP, bool) {
```
johanbrandhorst, line 18: What do you think about supporting https://pkg.go.dev/net/netip#Addr? I know `peer.FromContext` returns a `net.IP` but it's sort of deprecated in favor of `netip.Addr`. It should be possible to use https://pkg.go.dev/net/netip#ParseAddr to parse from net.IP.

johanbrandhorst (commented):
This is an _excellent_ PR, thank you! Just a few minor suggestions, let me know what you think.

interceptors/realip/examples_test.go
MadsRC: I've added them as package headers now. I originally didn't do it as to not let the package dictate what headers people should use. But I suppose making the well-known defaults available wouldn't hurt.

interceptors/realip/realip.go
MadsRC: Great catch!

interceptors/realip/realip.go
MadsRC: I've refactored to use netip.Addr ;)

MadsRC:
> This is an _excellent_ PR, thank you! Just a few minor suggestions, let me know what you think.

Thank you! I've implemented the suggestions you had. Turns out the `netip.Addr` rewrite was a bit more involved, but in the end it turned out great. I also rewrote the structure of the tests.

johanbrandhorst (changes requested):
This looks great, just some more minor points.

interceptors/realip/realip.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+// X_REAL_IP, X_FORWARDED_FOR and TRUE_CLIENT_IP are header keys
+// used to extract the real client IP from the request. They represent common
+// conventions for identifying the originating IP address of a client connecting
+// through proxies or load balancers.
+const (
+	X_REAL_IP       = "x-real-ip"
+	X_FORWARDED_FOR = "x-forwarded-for"
+	TRUE_CLIENT_IP  = "true-client-ip"
```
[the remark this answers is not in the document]
MadsRC: Absolutely - oversight on my part.

interceptors/realip/realip.go
```diff
@@ -0,0 +1,135 @@
...
+
+func getRemoteIP(ctx context.Context, trustedPeers []netip.Prefix, headers []string) netip.Addr {
+	pr := remotePeer(ctx)
+	if pr == nil {
+		return noIP
+	}
+	ip, err := netip.ParseAddr(strings.Split(pr.String(), ":")[0])
+	if err != nil {
+		return noIP
+	}
```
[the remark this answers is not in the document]
johanbrandhorst: Ah, my bad, thanks!

interceptors/realip/realip.go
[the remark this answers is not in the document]
johanbrandhorst: Sorry to bang on, but I don't see that the case of the headers was canonicalized?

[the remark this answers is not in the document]
MadsRC: Hey @johanbrandhorst - I guess I must have misunderstood the meaning of canonicalising the header keys. Let me go ahead and try to fix them ;)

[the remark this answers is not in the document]
MadsRC: I've changed the casing of the header keys that are the values of the constants - Is that what you meant by canonicalising the header keys?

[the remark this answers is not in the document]
johanbrandhorst: Yes, that is what I meant. I see you had to lowercase the headers in the map, which is fine, you could also canonicalize them using https://pkg.go.dev/net/textproto#CanonicalMIMEHeaderKey, but this will do.

johanbrandhorst (approved):
LGTM

johanbrandhorst:
Thanks for your contribution!

PR #684, opened 2023-12-20
protovalidate: add option to ignore certain message types

igor-tsiglyar:
## Changes

Add an option to ignore certain types of messages that are not ready to be validated at the middleware level and require some further processing and manual `protovalidate.Validate()` call.

## Verification

Unit test to verify that bad message can be ignored because of the option implemented.

### DIFF
--- a/interceptors/protovalidate/options.go
+++ b/interceptors/protovalidate/options.go
@@ -15,6 +15,7 @@ type options struct {
 	ignoreMessages []protoreflect.MessageType
 }
 
+// An Option lets you add options to protovalidate interceptors using With* funcs.
 type Option func(*options)
 
 func evaluateOpts(opts []Option) *options {
@@ -25,8 +26,8 @@ func evaluateOpts(opts []Option) *options {
 	return optCopy
 }
 
-// WithIgnoreMessages sets the messages that should be ignored as they are not yet ready
-// for validation at the stage this middleware operates
+// WithIgnoreMessages sets the messages that should be ignored by the validator. Use with
+// caution and ensure validation is performed elsewhere.
 func WithIgnoreMessages(msgs ...protoreflect.MessageType) Option {
 	return func(o *options) {
 		o.ignoreMessages = msgs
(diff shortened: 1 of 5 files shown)


interceptors/protovalidate/options.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"golang.org/x/exp/slices"
+	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+type options struct {
+	ignoreMessages []protoreflect.MessageType
+}
+
+type Option func(*options)
```
johanbrandhorst, line 18: Could you add a comment to this type please?

interceptors/protovalidate/options.go
```diff
@@ -0,0 +1,40 @@
...
+func evaluateOpts(opts []Option) *options {
+	optCopy := &options{}
+	for _, o := range opts {
+		o(optCopy)
+	}
+	return optCopy
+}
+
+// WithIgnoreMessages sets the messages that should be ignored as they are not yet ready
+// for validation at the stage this middleware operates
```
johanbrandhorst, line 29: I don't think we need to be this prescriptive about the use, we can just say
```suggestion
// WithIgnoreMessages sets the messages that should be ignored by the validator. Use with
// caution and ensure validation is performed elsewhere.
```

johanbrandhorst (changes requested):
Could you add an example_test.go file that shows how to use this new option? Thanks!

interceptors/protovalidate/options.go
igor-tsiglyar: done

interceptors/protovalidate/options.go
igor-tsiglyar: done

igor-tsiglyar:
> Could you add an example_test.go file that shows how to use this new option? Thanks!

@johanbrandhorst adjusted the existing interceptors/protovalidate/example_unary_test.go as I didn't see any usages of it, figured it is only for demonstrational purposes

also, handled the option in a stream handler for consistency

johanbrandhorst (approved):
LGTM!

johanbrandhorst:
Thanks for your contribution!

igor-tsiglyar:
@johanbrandhorst thanks

PR #687, opened 2024-01-12
Support retriable func condition

### LINKED ISSUE #686: Support extensibility of retry condition
Currently grpc_retry supports retry condition by setting `code.Codes`.
But it is good if retry condition is defined by error message, status details, etc...

So I propose below options.

```go
// When returns true, it is retriable.
type RetriableFunc func(err error) bool

// WithRetriable sets which error should be retried.
// When this option is set, WithCodes is ignored.
func WithRetriable(ErrorFunc) CallOption
```

tamayika:
Closes #686

## Changes

Add retry condition on arbitratry error.
This option overcedes WithCodes and default retriable logic.

## Verification

Test code added.

### DIFF
--- a/interceptors/retry/options.go
+++ b/interceptors/retry/options.go
@@ -9,6 +9,7 @@ import (
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 var (
@@ -22,11 +23,11 @@ var (
 		max:            0, // disabled
 		perCallTimeout: 0, // disabled
 		includeHeader:  true,
-		codes:          DefaultRetriableCodes,
 		backoffFunc:    BackoffLinearWithJitter(50*time.Millisecond /*jitter*/, 0.10),
 		onRetryCallback: OnRetryCallback(func(ctx context.Context, attempt uint, err error) {
 			logTrace(ctx, "grpc_retry attempt: %d, backoff for %v", attempt, err)
 		}),
+		retriableFunc: newRetriableFuncForCodes(DefaultRetriableCodes),
 	}
 )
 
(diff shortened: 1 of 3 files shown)


ggilmore:
This is a drive-by comment, but I think your approach would be better if you:

1. Have your `type RetriableFunc func(err error) bool` be the only option to be able to accept / reject errors (e.g. get rid of the status codes slice) from the retry `options` struct. 


Ex:

```diff
type options struct {
	max             uint
	perCallTimeout  time.Duration
	includeHeader   bool
-	codes           []codes.Code
	backoffFunc     BackoffFunc
	onRetryCallback OnRetryCallback
+	retriableFunc   RetriableFunc
}

```

2. Refactor the `WithCodes` options func as a `retriableFunc`:

```go
// WithCodes sets which codes should be retried.
//
// Please *use with care*, as you may be retrying non-idempotent calls.
//
// You cannot automatically retry on Cancelled and Deadline, please use `WithPerRetryTimeout` for these.
func WithCodes(retryCodes ...codes.Code) CallOption {
	return CallOption{applyFunc: func(o *options) {
                o.retriableFunc = func(err error) bool {
                     c := status.Code(err)

                     for _, code := retryCodes {
                        if c == code {
                           return true 
                        }
                     }
                     
                     return false
                 }
	}}
}
```

This maintains backwards compatibility with the existing of the library that are using `WithCodes`, and gives a library a deeper interface (fewer APIs needed to achieve the same functionality).

johanbrandhorst (changes requested):
Hi, thanks for your PR! This is nice, but I also agree with @ggilmore that we could simplify the internal logic to use the new function. Mind making the changes suggested?

tamayika:
@johanbrandhorst @ggilmore 
thanks, I refactored WithCodes option.

johanbrandhorst (approved):
LGTM

johanbrandhorst:
Thanks for your contribution!

PR #689, opened 2024-01-25
Fix `InitializeMetrics` signature to allow use with `xds.GRPCServer`

bozaro:
## Changes

Use interface instead of concrete type for `InitializeMetrics` signature.


johanbrandhorst (approved):
LGTM, though this is technically a breaking change, I doubt it will affect many users.

johanbrandhorst:
Thanks for your contribution!

PR #692, opened 2024-02-01
Extend realip parsing of GRPC peer address to handle IPv6

surik:
## Changes

Use `netip.ParseAddrPort` instead of `strings.Split` to support extraction of both ipv4 and ipv6.

## Verification

The `interceptors/realip/realip_test.go` file was extended with additional cases.


johanbrandhorst (approved):
LGTM

johanbrandhorst:
Thanks for your contribution!

PR #695, opened 2024-02-06
Extent realip interceptors with ip selection based on proxy count and list

surik:
The rightmost IP is not always the client IP. One example is Google:
https://cloud.google.com/load-balancing/docs/https#x-forwarded-for_header

The PR extends the IP selection for `X-Forwarded-For` based on [MDN Selecting an IP address](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#selecting_an_ip_address).

Just so you know, it is possible to configure both at the same time. 
The user needs to be cautious when configuring these for IP selection and normally has to pick `TrustedProxies` or `TrustedProxiesCount`.

## Changes

Two new methods `UnaryServerInterceptorOpts` and `StreamServerInterceptorOpts` and options:
```go
type Opts struct {
	TrustedPeers []netip.Prefix
	TrustedProxies []netip.Prefix
	TrustedProxiesCount uint
	Headers []string
}
```

The `UnaryServerInterceptor` and `StreamServerInterceptor` are kept for backward compatibility.

## Verification

The `realip_test.go` is extended with additional test cases.

### DIFF
--- a/interceptors/realip/realip.go
+++ b/interceptors/realip/realip.go
@@ -130,25 +130,12 @@ func (s *serverStream) Context() context.Context {
 	return s.ctx
 }
 
-// Opts specifies the options of the middleware
-type Opts struct {
-	// List of trusted peers network prefix. We will look at the headers from their requests
-	TrustedPeers []netip.Prefix
-	// List of trusted proxies network prefixes.
-	// When going through X-Forwarded-For the first rightmost non matching ip is the client ip.
-	TrustedProxies []netip.Prefix
-	// Number of proxies in front that may append X-Forwarded-For. 0 by default
-	TrustedProxiesCount uint
-	// Headers to use in real ip extraction when the request is from trusted peer.
-	Headers []string
-}
-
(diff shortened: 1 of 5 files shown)


interceptors/realip/realip.go
```diff
@@ -108,25 +130,56 @@ func (s *serverStream) Context() context.Context {
 	return s.ctx
 }
 
+// Opts specifies the options of the middleware
+type Opts struct {
```
johanbrandhorst, line 134: This is great, but we generally use functional options to configure the interceptors. Do you mind rewriting this in the functional option style? Thanks!

johanbrandhorst (changes requested):

interceptors/realip/realip.go
surik: Oh indeed, I had to look at this before. Now this is reworked to functional options.

johanbrandhorst (approved):
LGTM

johanbrandhorst:
Thanks for your contribution!

PR #700, opened 2024-03-12
Call retry callback on retry

### LINKED ISSUE #699: Use retry hooks for adding retry metrics
I want to add a metric (in v2) and log that shows how many retries are being made from the grpc interceptor, and from what I can tell there are two options, `WithOnRetryCallback` or `WithRetriable`, but both of them have some drawbacks:

- `WithRetryable`: The func in WithRetryable only gets the error, so tracking number of retries is not possible using this approach. 
- `WithOnRetryCallback`: WithOnRetryCallback is not actually called on retries, but rather whenever a call fails, even if that call will not be retried. So if I wanted a metric to show the number of actual retries, I would have to duplicate the code that check if it should retry.

I could achieve this by using both of the options, as I would then break out the is retriable check into my own code, and could use it to check in the on retry callback if it will actually retry, but this feels a bit hacky to me. 

I wonder if there could be an other approach to solve this. I had expected the on retry callback to actually only trigger on retries for example. Alternatively the context and attempt number could be passed to the retriable function.

Would you be open to such a change, or do you have any other ideas around this?

fredr:
## Changes

Call the on retry callback on retry rather than on failure.
Fixes #699 

This will change so that the callback will only be called when retrying, and never with attempt = 0 (because the first attempt is not a retry). Previously it was called when any attempt failed, even if there where no retries.

## Verification

I have tested this patch in our local environment where I've performed a load test to provoke some retries, and it now get triggered as I would expect, only on retries.

### DIFF
--- a/interceptors/retry/retry.go
+++ b/interceptors/retry/retry.go
@@ -93,6 +93,9 @@ func StreamClientInterceptor(optFuncs ...CallOption) grpc.StreamClientIntercepto
 
 		var lastErr error
 		for attempt := uint(0); attempt < callOpts.max; attempt++ {
+			if attempt > 0 {
+				callOpts.onRetryCallback(parentCtx, attempt, lastErr)
+			}
 			if err := waitRetryBackoff(attempt, parentCtx, callOpts); err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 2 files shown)


johanbrandhorst (changes requested):
Thanks for the fix, do you think you could modify the tests to assert this new behavior too? Thanks!

fredr:
NP, Thanks for reviewing! I've added a test for checking that the callback doesn't run for non retriable errors, is there anything else you wound want me to add a case for?

johanbrandhorst (approved):
LGTM!

PR #702, opened 2024-04-08
interceptors: Update logging interceptor Reporter to re-extract fields from context before logging

chancez:
When using logging.WithFieldsFromContext, if the value being extracted as a log field is modified after the logging interceptor initializes the Reporter before the underlying handler is called, then the updated value will not be reflected in the log message.

To fix this, re-extract fields from the context before logging them in PostCall, PostMsgSend and PostMsgReceive, ensuring the updated values in the context are logged.

## Changes

- Call `c.opts.fieldsFromCtxCallMetaFn` `PostCall`, `PostMsgSend` and `PostMsgReceive`
- Update tests to verify log fields derived from context values are updated in the `reporter.Post*` methods

## Verification

- Unit tests
- Used this branch in a (private) project where I was expecting this behavior to work, and after the patch is applied, the expected behavior is observed.

### DIFF
--- a/interceptors/logging/interceptors.go
+++ b/interceptors/logging/interceptors.go
@@ -41,6 +41,10 @@ func (c *reporter) PostCall(err error, duration time.Duration) {
 	if err != nil {
 		fields = fields.AppendUnique(Fields{"grpc.error", fmt.Sprintf("%v", err)})
 	}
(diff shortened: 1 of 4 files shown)


johanbrandhorst (approved):
Looks reasonable to me, thanks for the PR!

chancez:
Wow thanks for the promptness!

PR #703, opened 2024-04-08
logging: Document correct WithFieldsFromContext/WithFieldsFromContextAndCallMeta usage

chancez:
## Changes

Updates the doc string for `logging.WithFieldsFromContext` and `logging.WithFieldsFromContextAndCallMeta`. I mistakenly thought these accumulated, so when I tried using one helper for logging some SQL query metadata _and_ the traceID, only the traceID was getting logged since it overwrote my previous field extractor func. Without reading the source, it's not entirely obvious this is the behavior. Additionally, `WithFieldsFromContext` and `WithFieldsFromContextAndCallMeta` are mutually exclusive, so it should be documented only one should be used, and what happens when you use both, or either one of them more than once.

### DIFF
--- a/interceptors/logging/options.go
+++ b/interceptors/logging/options.go
@@ -138,7 +138,11 @@ type (
 	fieldsFromCtxCallMetaFn func(ctx context.Context, c interceptors.CallMeta) Fields
 )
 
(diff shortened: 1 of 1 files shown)


johanbrandhorst (approved):
Nice, thank you!

PR #708, opened 2024-04-25
feat(auth): add partial per service auth override

### LINKED ISSUE #707: Feature: Auth Interceptor: Add support for partial per service AuthFunc override
Sometimes, there is a possibility that some of the methods in a service are publicly available and some need authentication.
The current ServiceAuthFuncOverride allows each service to override the global AuthFunc. However, if a service needs to override the global AuthFunc for some methods, it needs to duplicate AuthFunc logic.
It would be nice to allow services to override AuthFunc, but also allow them to fallback to the global AuthFunc if they want to only override it for certain methods and not all of them.

seinshah:
closes #707 

## Changes

1. Added public `ErrNoAuthOverrideMatch` to `auth` package
2. Continue to run the provided `AuthFunc` when `AuthFuncOverride` returns `ErrNoAuthOverrideMatch`
3. Added test cases

## Verification

- tested in my private project
- added test case to ensure correctness when `AuthFuncOverride` returns `ErrNoAuthOverrideMatch`.


johanbrandhorst:
Hi, thanks for the PR. Please don't submit PRs until we've agreed on a way forward to an issue.

seinshah:
@johanbrandhorst, in case of approving the issue, I wanted the issue to be on a fast track. It might be good to add this guideline to `CONTRIBUTING.md` file as well to prevent future cases. Thanks.

PR #711, opened 2024-05-07
protovalidate: Add custom ErrorConverter 

khasanovbi:
## Changes

- Fix validation for grpc stream, validate before message handling.
- Add WithErrorConverter option to customize protovalidate output.

## Verification

- Add tests to related changes.
- Check in my production.

### DIFF
--- a/interceptors/protovalidate/protovalidate.go
+++ b/interceptors/protovalidate/protovalidate.go
@@ -5,34 +5,38 @@ package protovalidate
 
 import (
 	"context"
-	"errors"
 
 	"github.com/bufbuild/protovalidate-go"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
 )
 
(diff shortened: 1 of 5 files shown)


johanbrandhorst:
Hi, could you please open an issue first to discuss this change? I'm not sure I understand/agree that this is necessary. Thanks!

interceptors/protovalidate/protovalidate.go
```diff
@@ -41,55 +45,40 @@ func UnaryServerInterceptor(validator *protovalidate.Validator, opts ...Option)
...
-	msg := m.(proto.Message)
-	if w.options.shouldIgnoreMessage(msg.ProtoReflect().Type()) {
-		return nil
-	}
-	if err := w.validator.Validate(msg); err != nil {
-		return status.Error(codes.InvalidArgument, err.Error())
-	}
-
-	return nil
+	return w.ServerStream.RecvMsg(m)
```
akshayjshah, line 67: Is this correct? From what I can tell, `RecvMsg` is unmarshalling data into `m`, so we should validate _after_ calling it.

PR #714, opened 2024-05-25
Include error details in protovalidate responses

akshayjshah:
This PR improves the protovalidate middleware:

- First, it removes a potential panic from the streaming interceptor.
- Second, it uses error details to return structured information to callers
  when message validation fails. This allows clients to know which field(s) of
  the request were invalid and which validation rules failed.

Speaking as one of the protovalidate authors, we intended validation errors to
be used as gRPC error details. I didn't notice that we weren't sending them
over the wire when I first reviewed this code.

### DIFF
--- a/interceptors/protovalidate/protovalidate_test.go
+++ b/interceptors/protovalidate/protovalidate_test.go
@@ -9,16 +9,19 @@ import (
 	"net"
 	"testing"
 
+	"buf.build/gen/go/bufbuild/protovalidate/protocolbuffers/go/buf/validate"
 	"github.com/bufbuild/protovalidate-go"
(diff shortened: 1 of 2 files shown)


interceptors/protovalidate/protovalidate_test.go
```diff
@@ -9,16 +9,19 @@ import (
 	"net"
 	"testing"
 
+	"buf.build/gen/go/bufbuild/protovalidate/protocolbuffers/go/buf/validate"
```
johanbrandhorst, line 12: _nice_ 😄

johanbrandhorst (approved):
Good to see you Akshay 😄. LGTM!

akshayjshah:
Thanks for the quick review, @johanbrandhorst! I also noticed another bug in the streaming interceptor, but I'll fix that in a separate PR.

PR #715, opened 2024-05-25
protovalidate: avoid pointer comparisons

akshayjshah:
## Changes
- Change the logic used in `WithIgnoreMessages` to avoid fragile pointer
comparisons. Instead, we match messages using their fully-qualified names.
- Mention error details in the interceptor docstrings.
- Make the tests more robust by asserting whether or not the service
  implementation should be called.

## Verification

The unit tests should fully exercise the changes made.

### DIFF
--- a/interceptors/protovalidate/options.go
+++ b/interceptors/protovalidate/options.go
@@ -12,7 +12,7 @@ import (
 )
 
 type options struct {
-	ignoreMessages []protoreflect.MessageType
+	ignoreMessages []protoreflect.FullName
 }
 
 // An Option lets you add options to protovalidate interceptors using With* funcs.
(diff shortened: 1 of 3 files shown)


interceptors/protovalidate/protovalidate.go
```diff
@@ -15,98 +15,76 @@ import (
...
 	}
-
-	return nil
-}
-
-// wrappedServerStream is a thin wrapper around grpc.ServerStream that allows modifying context.
-type wrappedServerStream struct {
-	grpc.ServerStream
-	// wrappedContext is the wrapper's own Context. You can assign it.
-	wrappedContext context.Context
```
akshayjshah, line 84: This field wasn't used anywhere, so I deleted it. 🤷🏽‍♂️

interceptors/protovalidate/options.go
```diff
@@ -26,16 +26,23 @@ func evaluateOpts(opts []Option) *options {
...
+		o.ignoreMessages = names
 	}
 }
 
-func (o *options) shouldIgnoreMessage(m protoreflect.MessageType) bool {
-	return slices.ContainsFunc(o.ignoreMessages, func(t protoreflect.MessageType) bool {
-		return m == t
-	})
+func (o *options) shouldIgnoreMessage(fqn protoreflect.FullName) bool {
+	_, found := slices.BinarySearch(o.ignoreMessages, fqn)
```
johanbrandhorst, line 46: 👌🏻 . Mind adding a comment here saying that we can use binary search since we're doing a sort above?

johanbrandhorst (changes requested):
Thanks, this looks great, mind adding some comments?

interceptors/protovalidate/options.go
akshayjshah: Done!

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

PR #716, opened 2024-05-25
Update protovalidate and grpc-go

akshayjshah:
## Changes

Update protovalidate and gRPC.

## Verification

Existing tests should be sufficient.


akshayjshah:
Yikes, I give up!

PR #721, opened 2024-08-14
Protovalidate interceptor cleanup, Go version bump

ash2k:
## Changes

Please see individual commits.

## Verification

Unit tests. No significant changes.


ash2k:
Follow up questions:

- I have a custom codec that I use with gRPC. It handles both proto and non-proto (a single type) messages. I'd like to use the protovalidate interceptors with it. Would you accept a feature to specify a list of Go types (as `reflect.Type`) to ignore, instead of erroring on them.
- Would you accept a PR with client validating interceptors?

interceptors/protovalidate/protovalidate.go
```diff
@@ -14,9 +14,13 @@ import (
...
 )
 
+type Validator interface {
+	Validate(proto.Message) error
+}
+
 // UnaryServerInterceptor returns a new unary server interceptor that validates incoming messages.
 // If the request is invalid, clients may access a structured representation of the validation failure as an error detail.
-func UnaryServerInterceptor(validator *protovalidate.Validator, opts ...Option) grpc.UnaryServerInterceptor {
+func UnaryServerInterceptor(validator Validator, opts ...Option) grpc.UnaryServerInterceptor {
```
johanbrandhorst, line 23: This is actually a breaking change, and I'm not sure we need to make this change? If you want to propose this, please open an issue. If you remove this change I'm happy to merge the rest of this PR.

johanbrandhorst (changes requested):

johanbrandhorst:
>     * I have a custom codec that I use with gRPC. It handles both proto and non-proto (a single type) messages. I'd like to use the protovalidate interceptors with it. Would you accept a feature to specify a list of Go types (as `reflect.Type`) to ignore, instead of erroring on them.

No, I'd rather users create their own interceptors for this sort of thing.

>     * Would you accept a PR with client validating interceptors?

For outgoing requests? For responses from servers? Not sure where that would go.

interceptors/protovalidate/protovalidate.go
ash2k: Removed this commit.

johanbrandhorst (approved):
LGTM

johanbrandhorst:
Thanks for your contribution!

ash2k:
Thank you!

PR #723, opened 2024-08-30
Use ValueFromIncomingContext() to reduce allocations and copying

ash2k:
## Changes

This function is more memory efficient as it doesn't copy all of the metadata key-values, only what is requested.

Docs https://pkg.go.dev/google.golang.org/grpc/metadata#ValueFromIncomingContext.

## Verification

Unit tests.

### DIFF
--- a/interceptors/auth/metadata.go
+++ b/interceptors/auth/metadata.go
@@ -7,12 +7,12 @@ import (
 	"context"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


johanbrandhorst (approved):
LGTM

johanbrandhorst:
Thanks for this!

PR #734, opened 2024-11-30
logging: add WithErrorFields

kindermoumoute:
> `WithErrorFields` allows to extract logging fields from an error.

Typically to add a stack trace field or more context around the error.

## Changes

- logging: add a new `WithErrorFields` option
- testpb: add `wrappedErrFields` and helpers to inject / extract fields
- testpb: PingError return a `wrappedErrFields`

## Verification

In PingError test (logging interceptor), assertion on finished fields show `"error-field"` is present and set to the correct value.


johanbrandhorst (approved):
Sure, sounds good to me!

PR #739, opened 2024-12-20
logging: add AddFields

kindermoumoute:
Support the similar helpers from the v1 introduced in #91:
- [ctxzap.AddFields](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/v1/logging/zap/ctxzap/context.go#L23-L32)
- [ctxlogrus.AddFields](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/v1/logging/logrus/ctxlogrus/context.go#L22-L30)
- [ctxkit.AddFields](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/v1/logging/kit/ctxkit/context.go#L22-L28)

Use case: let an interceptor or the grpc method implemention add more fields to logging interceptor (e.g. "finished call").

## Changes

- store the fields into the `fieldsCtxValue` struct
- add function `AddFields`
- add test `TestAddFields`

## Verification

Old tests and new tests pass.

### DIFF
--- a/interceptors/logging/logging.go
+++ b/interceptors/logging/logging.go
@@ -191,6 +191,9 @@ func InjectLogField(ctx context.Context, key string, val any) context.Context {
 // AddFields updates the fields already in the context that will be used by the logging interceptor or can be
 // extracted further in ExtractFields. For explicitness, in case of duplicates, the newest field occurrence wins.
 //
+// InjectFields should be used instead of AddFields where possible, as it does not require mutating the values
+// already in the context.
+//
 // This function is not safe to call concurrently.
 func AddFields(ctx context.Context, f Fields) {
 	t, ok := ctx.Value(fieldsCtxMarkerKey).(*fieldsCtxValue)
--- a/interceptors/logging/logging.go
+++ b/interceptors/logging/logging.go
@@ -22,6 +22,9 @@ var (
 )
 
 type fieldsCtxMarker struct{}
+type fieldsCtxValue struct {
+	fields Fields
+}
 
 var (
 	// fieldsCtxMarkerKey is the Context value marker that is used by logging middleware to read and write logging fields into context.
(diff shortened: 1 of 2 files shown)


interceptors/logging/logging.go
```diff
@@ -177,14 +180,25 @@ func ExtractFields(ctx context.Context) Fields {
 //
 // Don't overuse mutation of fields to avoid surprises.
 func InjectFields(ctx context.Context, f Fields) context.Context {
-	return context.WithValue(ctx, fieldsCtxMarkerKey, f.WithUnique(ExtractFields(ctx)))
+	return context.WithValue(ctx, fieldsCtxMarkerKey, &fieldsCtxValue{f.WithUnique(ExtractFields(ctx))})
```
johanbrandhorst, line 183: Mind using explicit field names?
```suggestion
	return context.WithValue(ctx, fieldsCtxMarkerKey, &fieldsCtxValue{fields: f.WithUnique(ExtractFields(ctx))})
```

johanbrandhorst (commented):
Could you clarify the difference between `AddFields` and `InjectFields`?

kindermoumoute:
> Could you clarify the difference between `AddFields` and `InjectFields`?

I updated the descriptions of both:
- **InjectFields** returns a new context with merged fields 
- **AddFields** updates the fields already in the context

johanbrandhorst:
> > Could you clarify the difference between `AddFields` and `InjectFields`?
> 
> I updated the descriptions of both:
> 
>     * **InjectFields** returns a new context with merged fields
> 
>     * **AddFields** updates the fields already in the context

Thanks for the explanation, but I'm a bit confused. This seems like an anti pattern to me. Contexts are generally immutable. Could you elaborate on the use case?

kindermoumoute:
> Thanks for the explanation, but I'm a bit confused. This seems like an anti pattern to me. Contexts are generally immutable. Could you elaborate on the use case?

The main use case is to add business-level attributes into the existing context so they appear in the final logs.

```go
func (s *server) GetFoo(ctx context.Context, _ *emptypb.Empty) (*bar.GetFooResponse, error) {
	// Retrieve some metadata from the database
	fooMetadata := model.GetFoo(ctx)

	// AddFields makes these metadata visible from the "finished call" log
	logging.AddField(ctx, "metadata", fooMetadata)

	return &bar.GetFooResponse{...}, nil
}
```

Another approach would be to generate another line of log at the business level, but the idea here is also to limit the amount of line of logs to 1. I agree it does sound anti-pattern to use context this way, but I couldn't find a better workaround yet - I think this is why AddFields was added and maintained in the first place in the v1.

interceptors/logging/logging.go
```diff
@@ -161,30 +164,42 @@ NextAddField:
...
 
 // InjectLogField is like InjectFields, just for one field.
 func InjectLogField(ctx context.Context, key string, val any) context.Context {
 	return InjectFields(ctx, Fields{key, val})
 }
 
+// AddFields updates the fields already in the context that will be used by the logging interceptor or can be
+// extracted further in ExtractFields. For explicitness, in case of duplicates, the newest field occurrence wins.
+//
+// This function is not safe to call concurrently.
```
johanbrandhorst, line 194: I'm sympathetic to the use case, but I do kinda hate the way we have to implement it. I don't feel strongly enough to refuse this functionality, but maybe we could add a note to the comment to say that in general, `InjectFields` should be preferred over `AddFields`?
```suggestion
// AddFields updates the fields already in the context that will be used by the logging interceptor or can be
// extracted further in ExtractFields. For explicitness, in case of duplicates, the newest field occurrence wins.
//
// InjectFields should be used instead of AddFields where possible, as it does not require mutating the values
// already in the context.
//
// This function is not safe to call concurrently.
```

johanbrandhorst (changes requested):

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contributino!

PR #740, opened 2024-12-27
logging: store the propagated context in the reporter

kindermoumoute:
https://github.com/grpc-ecosystem/go-grpc-middleware/pull/739 doesn't work without this change because the fields were not extracted from the same context than the one being propagated. I should've tested this better in my previous MR 🙌

## Changes

- Use a context with injected field in the reporter in order to support https://github.com/grpc-ecosystem/go-grpc-middleware/pull/739
- Add a helper to execute code without ciruclar dependency from the `testpb.TestPingService.Ping` method

## Verification

- Call `AddFields` with the `Ping` method using the `PingFunc` helper
- Update `TestPing` to assert a new field


johanbrandhorst (approved):

johanbrandhorst:
Thank you for your contribution!

PR #760, opened 2025-06-18
fix: correct comment typo in StreamServerInterceptor function

haru-256:
## Changes

- fix typo for StreamServerInterceptor

## Verification

run the following command
- `make all`

### DIFF
--- a/interceptors/auth/auth.go
+++ b/interceptors/auth/auth.go
@@ -49,7 +49,7 @@ func UnaryServerInterceptor(authFunc AuthFunc) grpc.UnaryServerInterceptor {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


johanbrandhorst (approved):
Thanks!

Oleja123:
Guys, is the comment correct? Looks like auth logic executes only on streaming creation, not per-request in stream

johanbrandhorst:
That's fair, the latter part of the comment could be updated, but that is a consequence of how gRPC stream interceptors work in general. Feel free to submit a PR updating it.

Oleja123:
> That's fair, the latter part of the comment could be updated, but that is a consequence of how gRPC stream interceptors work in general. Feel free to submit a PR updating it.

Ok, thanks! I will do this

PR #761, opened 2025-06-22
Minor refactor to remove duplication

nwnt:
Minor refactoring to remove duplication found in the logging interceptors.

## Changes

Extract the only difference between the if and else blocks.

## Verification

`go test` in the package and then `make test` in the project root dir.


johanbrandhorst (approved):
Sure, seems simple enough

johanbrandhorst:
Thanks for your contribution :)

PR #763, opened 2025-07-17
Added WithIgnoreUnknownType option to protovalidate interceptor

InbarPerry:
# Add WithIgnoreUnknownType option to protovalidate interceptor

## Problem

Protovalidate interceptors fail when used with transparent gRPC proxies that use custom codecs. These proxies handle raw bytes/frames instead of concrete protobuf types, causing `"unsupported message type"` errors.

## Solution

Added `WithIgnoreUnknownType()` option that allows non-protobuf messages to pass through validation while still validating actual protobuf messages.

```go
// Before: FAILS with proxy codecs
server := grpc.NewServer(
    grpc.CustomCodec(transparentCodec),
    grpc.StreamInterceptor(protovalidate.StreamServerInterceptor(validator)), // Error here
    grpc.UnknownServiceHandler(proxyHandler),
)

// After: Works with proxy codecs
server := grpc.NewServer(
    grpc.CustomCodec(transparentCodec),
    grpc.StreamInterceptor(protovalidate.StreamServerInterceptor(validator,
        protovalidate.WithIgnoreUnknownType(), // Allows compatibility
    )),
    grpc.UnknownServiceHandler(proxyHandler), // Transparent forwarding
)
```
## Changes

- Add WithIgnoreUnknownType()` in `options.go`
- Modified `validateMsg()` to skip type checking when option is enabled

## Verification
- Added tests for both unary and streaming interceptors that checks the option with and without the existing `WithIgnoreMessages` option.

### DIFF
--- a/interceptors/protovalidate/options.go
+++ b/interceptors/protovalidate/options.go
@@ -13,7 +13,8 @@ import (
 )
 
 type options struct {
-	ignoreMessages []protoreflect.FullName
+	ignoreMessages    []protoreflect.FullName
+	ignoreUnknownType bool
 }
 
 // An Option lets you add options to protovalidate interceptors using With* funcs.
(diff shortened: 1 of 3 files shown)


johanbrandhorst:
Hi, I'm not convinced this is functionality we need in this library. It seems a very niche use case. I often say that users should just fork the library if they have special uses. Every new option adds a cost to the maintenance and the use of the library, so new features need to be worth the cost. I also prefer if you open an issue first before just creating a PR so we can discuss the merit of the feature before anyone invests time building it.

PR #786, opened 2025-08-05
chore: enable hugeParam rule from go-critic

mmorel-35:
## Changes

Enable [hugeParam](https://go-critic.com/overview.html#hugeparam) rule from go-critic


johanbrandhorst:
Thanks!

Issue #794, opened 2025-08-14
TestServiceClient.PingError causes panic


floppyzedolfin:
Hello,
I'm testing a personal interceptor, and I realised that my test was crashing because the `PingError` method of the `TestServiceClient` panics in some conditions.

Here is the code I use to reproduce the issue:

```go
package interceptor_test

import (
	"net"
	"testing"

	"github.com/grpc-ecosystem/go-grpc-middleware/v2/testing/testpb"
	"google.golang.org/grpc"
	"google.golang.org/grpc/credentials/insecure"
)

func TestInterceptor(t *testing.T) {
	lstConfig := net.ListenConfig{}

	serverListener, err := lstConfig.Listen(t.Context(), "tcp", "")
	if err != nil {
		t.Fatalf("failed to listen: %v", err)
	}

	srv := grpc.NewServer()
	testpb.RegisterTestServiceServer(srv, &testpb.TestPingService{})

	go func() {
		err := srv.Serve(serverListener)
		if err != nil {
			panic(err)
		}
	}()

	grpcClientConn, err := grpc.NewClient(
		serverListener.Addr().String(),
		grpc.WithTransportCredentials(insecure.NewCredentials()),
	)
	if err != nil {
		t.Fatalf("failed to connect: %v", err)
	}

	cli := testpb.NewTestServiceClient(grpcClientConn)

	defer func() {
		srv.Stop()
	}()

	_, err = cli.PingError(t.Context(), &testpb.PingErrorRequest{
		Value: "hello",
	})
	if err != nil {
		t.Fatal("calling Ping Err:", err.Error())
	}
}
```

Here is the output of running this test:
```
▶ go test ./...
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x8d6b35]

goroutine 22 [running]:
github.com/grpc-ecosystem/go-grpc-middleware/v2/testing/testpb.(*wrappedErrFields).Error(0xc0002146c0?)
	/home/pascal/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-middleware/v2@v2.3.2/testing/testpb/interceptor_suite.go:164 +0x15
google.golang.org/grpc/status.FromError({0xb3b2e0, 0xc0002146c0})
	/home/pascal/go/pkg/mod/google.golang.org/grpc@v1.74.2/status/status.go:108 +0xd3
[...]

PR #795, opened 2025-08-18
fix(#794): Wrapping codes.OK should not cause panic

floppyzedolfin:
An panic was raised in `wrappedErrField.Error()` when `PingError` was used without specifying the `ErrorCodeReturned`.
See #794 for a description of the original issue, with context.
It turns out it is possible to build a `wrappedErrFields` which has a `wrappedErr == nil`.

## Changes

Added some safety nets in the calls accessing the `err.wrappedErr` field.

## Verification

Added a unit test that covers the "usual" scenario (when the `ErrorCodeReturned` was properly set) and the edge-case scenario (when no `ErrorCodeReturned` was set).


johanbrandhorst (approved):
Thanks!

PR #809, opened 2025-10-29
avoid unnecessary logging field creation when payload logging is disabled

dbeneker:
While debugging a performance-critical gRPC stream, I noticed that the logging reporter creates logging fields for every sent and received message, even when the `LoggableEvent` options `PayloadReceived` and `PayloadSent` are not set.
This unnecessary field creation increases garbage collection pressure and negatively impacts throughput.

## Changes

Added an early return in the `PostMsgSend()` and `PostMsgReceive()` to skip building logging fields when payload logging is disabled.

## Verification

Verified with existing unit tests.


dbeneker:
Tanks for merging! I noticed that the latest release tag hasn’t been updated in quite a while. Could you create a new tag?

Issue #804, opened 2025-09-25
[Prometheus] grpc_server_handled_total incorrectly registers context labels as grpc_code during pre registration


UN0wen:
**Issue**
When using grpc server metrics and using `WithContextLabels`, upon preregistration, the first context label is given values that should go into the `grpc_code` label instead

```golang
package grpcserver

import (
   grpcprom "github.com/grpc-ecosystem/go-grpc-middleware/providers/prometheus"
  "github.com/prometheus/client_golang/prometheus"
  "google.golang.org/grpc"
)

serverMetrics := grpcprom.NewServerMetrics(
	grpcprom.WithServerHandlingTimeHistogram(
		grpcprom.WithHistogramBuckets(prometheus.DefaultGRPCServerDurationBuckets),
	),
	grpcprom.WithContextLabels("environment"),
        grpcprom.WithContextLabels("context-label-2"),
)

// labels from context etc.

prometheus.MustRegister(serverMetrics)

grpcServer := grpc.NewServer(
   grpc.ChainUnaryInterceptor(
      serverMetrics.UnaryServerInterceptor()
   )
)

// gprc service registration 

serverMetrics.InitializeMetrics(grpcServer)
```

What I expect after `InitializeMetrics`:
```
grpc_server_handled_total{grpc_method="...", grpc_service="...", grpc_type="unary", grpc_code="Aborted"}
```

What I got:
```
grpc_server_handled_total{grpc_method="...", grpc_service="...", grpc_type="unary", environment="Aborted"}
```

I checked the code and it looks like it's just a minor ordering issue [during preregistration](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/main/providers/prometheus/server_metrics.go#L144). Will look into contributing a fix for this once I have time.

PR #810, opened 2025-10-30
[Prometheus] Fix pre-registration of `handled` metrics with context labels

t-bowcock:
## Changes

Build label list `handledLabelsWithCode` in loop to place `code.String()` in correct place.
Fixes https://github.com/grpc-ecosystem/go-grpc-middleware/issues/804

## Verification

Tested with grpc server on a personal project.


johanbrandhorst (approved):
I wouldn't worry about that test failure, there's no way this is related. Thanks for the fix!

